PR #3, opened 2023-07-28
Add nmconnection files from the referenced configmap

carbonin:
Resolves https://issues.redhat.com/browse/MGMT-15245

cc @eranco74

### DIFF
--- a/config/crd/bases/relocation.openshift.io_clusterconfigs.yaml
+++ b/config/crd/bases/relocation.openshift.io_clusterconfigs.yaml
@@ -371,7 +371,9 @@ spec:
                 x-kubernetes-map-type: atomic
               networkConfigRef:
                 description: NetworkConfigRef is the reference to a config map containing
-                  network configuration files if necessary
+                  network configuration files if necessary. Keys should be of the
+                  form *.nmconnection and each represent an nmconnection file to be
+                  applied to the host.
                 properties:
                   name:
                     description: 'Name of the referent. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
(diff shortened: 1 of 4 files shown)


config/crd/bases/relocation.openshift.io_clusterconfigs.yaml
```diff
@@ -162,7 +162,9 @@ spec:
                 x-kubernetes-map-type: atomic
               networkConfigRef:
                 description: NetworkConfigRef is the reference to a config map containing
-                  network configuration files if necessary
+                  network configuration files if necessary. Keys should be of the
```
eranco74, line 165: OK, this isn't the same configuration assisted get.
Assisted is getting nmstate policy files and generates the nmconnection from it (by executing `nmstatectl gc` command).
Here the API asks for a nmconnection file.
While we want to keep this operator simple there's an advantage of keeping the same API as ZTP.
Also, in the feuture (I think it's targeting 4.14) coreos should have nmstate lib installed and nmstate.service enabled, so we could just place the policy file under /etc/nmstate and the nmstate service will just take care of it (we won't need to generate the nmconnection file)

eranco74 (commented):
Unsure if we want nmconnection file or nm policy file as input

config/crd/bases/relocation.openshift.io_clusterconfigs.yaml
carbonin: >we won't need to generate the nmconnection file

This is what I was trying to avoid because we had a bunch of issues getting it to work correctly.

So what are you thinking? Just a configmap but assume the contents are nmstate?
Will we need more than one file?

eranco74: I don't know if we can rely on having nmstate.service.
I also don't want to start generating the connection files in this service.
I guess we can ask for nmconnection files and change the API if we must.
And yes, we might need more than one file, the nodes might have few nics that get's static network configs.

carbonin: Okay, I'll leave this as is for now and we could always support both in the future if we need to.

PR #4, opened 2023-07-28
Label ISO with 'ZTC SNO'

carbonin:
The test uses `HavePrefix` as checking for exact equality leads to the following failure:

```
  [FAILED] Expected
      <string>: ZTC SNO\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00
  to equal
      <string>: ZTC SNO
```

cc @eranco74 

I just opened this to make this agree with the current state of https://github.com/eranco74/sno-relocation-poc/blob/master/bake/installation-configuration.sh#L26, it's up to you if you want to change it here or there.

### DIFF
--- a/internal/imageserver/imageserver.go
+++ b/internal/imageserver/imageserver.go
@@ -75,7 +75,7 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		w.WriteHeader(http.StatusInternalServerError)
 		return
 	}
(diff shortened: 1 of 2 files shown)


eranco74:
I'll change it on my end, `relocation-config` seems better than ZTC-SNO

eranco74:
Done
https://github.com/eranco74/sno-relocation-poc/commit/94c5abce96593c46857958946bfa433daa893ff4

PR #5, opened 2023-07-31
MGMT-15407: Add ACM registration information to ISO

carbonin:
Update cluster relocation operator to v0.9.7 to bring in ACM registration data and add the referenced secret to the ISO payload.

Resolves https://issues.redhat.com/browse/MGMT-15407


eranco74 (commented):
/lgtm
Note that like the pull secret, the acm secret need to be added to the json/yaml in the ISO

controllers/clusterconfig_controller.go
```diff
@@ -226,6 +226,12 @@ func (r *ClusterConfigReconciler) writeInputData(ctx context.Context, config *re
 			return fmt.Errorf("failed to write pull secret: %w", err)
 		}
 
+		if config.Spec.ACMRegistration != nil {
+			if err := r.writeSecretToFile(ctx, &config.Spec.ACMRegistration.ACMSecret, filepath.Join(filesDir, "acm-secret.json")); err != nil {
```
carbonin, line 230: >Note that like the pull secret, the acm secret need to be added to the json/yaml in the ISO

@eranco74 that's what we're doing here.

Not sure if the comment was just for information or if you thought this part wasn't done yet.

PR #7, opened 2023-08-03
MGMT-15282: Remove files and deprovision BMH when clusterconfig is removed

carbonin:
Adds a finalizer to clusterconfig resources that allows the controller
to clean up the local files for that config and remove the image
reference from a linked BMH.

Resolves https://issues.redhat.com/browse/MGMT-15282


eranco74 (commented):
/lgtm

PR #9, opened 2023-08-03
Make kustomize image substitution work properly

carbonin:
This is required for `make deploy` to properly substitute the image using the `IMG` env var.

Without this `quay.io/carbonin/cluster-relocation-service:latest` is used unconditionally.

cc @achuzhoy


eranco74 (commented):
/lgtm

PR #10, opened 2023-08-04
Serve images over https

carbonin:
Configure service serving certs via annotation and mount the resulting secret into the service container

Resolves https://issues.redhat.com/browse/MGMT-15257


eranco74 (commented):
/lgtm

PR #11, opened 2023-08-04
Detach BMH after it is provisioned

carbonin:
This prevents the BMH from accidentally being used later and rebooting the host and from the host rebooting when the clusterconfig is removed.

Resolves https://issues.redhat.com/browse/MGMT-15247


eranco74 (commented):
/lgtm

PR #12, opened 2023-08-07
Add namespace and use it for all resources

carbonin:
Before this change this service was not providing a manifest for the target namespace of any of the resources. Additionally the referenced secret could be in namespaces separate from the CluterConfig.

This commit creates a single namespace manifest and uses that namespace for all of the created resources (clusterrelocation and secrets).

This will make the resulting resources on the relocated cluster easier to find and understand when multiple clusters are being relocated using the same hub.

Resolves https://issues.redhat.com/browse/MGMT-15488

### DIFF
--- a/controllers/clusterconfig_controller.go
+++ b/controllers/clusterconfig_controller.go
@@ -63,6 +63,7 @@ type ClusterConfigReconciler struct {
 const (
 	detachedAnnotation    = "baremetalhost.metal3.io/detached"
 	clusterRelocationName = "cluster"
+	relocationNamespace   = "cluster-relocation"
 )
 
 //+kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch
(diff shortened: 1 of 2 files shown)


controllers/clusterconfig_controller.go
```diff
@@ -59,7 +59,10 @@ type ClusterConfigReconciler struct {
 	BaseURL string
 }
 
-const detachedAnnotation = "baremetalhost.metal3.io/detached"
+const (
+	detachedAnnotation  = "baremetalhost.metal3.io/detached"
+	relocationNamespace = "cluster-relocation"
```
eranco74, line 64: Why do we need this?
Can't we keep the resources with the same namespace as they have on the HUB?

carbonin: This would mean every relocated cluster from the same hub would have a different namespace.

Also the secrets can be in a different namespace from the cluster relocation resource so it seemed easier to change everything to one rather than keep track of several namespaces that I need to create.

carbonin: Is there any downside to creating a single namespace in the relocated cluster?

eranco74: No downside, just wonered why not keeping it simple, seems like good enough reasons

PR #13, opened 2023-08-07
Allow providing arbitrary additional manifests

carbonin:
ExtraManifestsRef is a reference to a config map containing additional
manifests to be applied to the relocated cluster.

Each key in the configmap will be considered a different file.
The manifests will be present in the `extra-manifests` directory in the
root of the ISO.

Resolves https://issues.redhat.com/browse/MGMT-15487

cc @imiller0 @browsell @eranco74


controllers/clusterconfig_controller.go
```diff
@@ -244,6 +244,25 @@ func (r *ClusterConfigReconciler) writeInputData(ctx context.Context, config *re
 			}
 		}
 
+		if config.Spec.ExtraManifestsRef != nil {
```
eranco74, line 247: Consider validating these files, at least the format.
Otherwise it might fail miserably later once we try to apply it on the cluster.

eranco74 (commented):
/lgtm

PR #14, opened 2023-08-08
Removing the option to only remove the CRD

achuzhoy:
This option can be confusing next to the option to remove the controller and clean the CRDs altogether.


achuzhoy:
cc @carbonin

carbonin (approved):

PR #21, opened 2023-12-14
Rework the API to use the lifecycle-agent (IBU) instead of cluster-relocation-operator

carbonin:
This removes the cluster-relocation-operator from this project and instead focuses on reusing the IBU (image based upgrade) work in https://github.com/openshift-kni/lifecycle-agent

Specifically it adds `ClusterInfo` instead of `ClusterRelocationSpec` in the API and explicitly handles writing out the pull secret and CA bundle.

cc @tsorya


api/v1alpha1/clusterconfig_types.go
```diff
@@ -36,20 +35,53 @@ const (
...
-	ExtraManifestsRef *corev1.LocalObjectReference `json:"extraManifestsRef,omitempty"`
+	ExtraManifestsRefs []corev1.LocalObjectReference `json:"extraManifestsRef,omitempty"`
+
+	// BareMetalHostRef identifies a BareMetalHost object to be used to attach the configuration to the host.
+	// +optional
+	BareMetalHostRef *BareMetalHostReference `json:"bareMetalHostRef,omitempty"`
+}
+
+// ClusterInfo struct that describe current cluster critical info
+type ClusterInfo struct {
```
tsorya, line 64: Maybe let me make it pkg in LCA?

carbonin: Yeah, I was debating an import, but didn't attempt it.

LCA doesn't actually seem that hard to import, is this exported there?

carbonin: Looks public. I can probably just try to import.

ref: https://github.com/openshift-kni/lifecycle-agent/blob/37291358c89d72ecf7b8ae1e37f752d901a729b9/ibu-imager/clusterinfo/clusterinfo.go#L30

tsorya: We should. If needed i will move it where you will need it :)

carbonin: Done

PR #22, opened 2024-01-04
MGMT-16431: Create admin kubeconfig

eranco74:
This PR adds the generation of the kubeconifg along with all required certificates for recert to configure the cluster with the given crypto.
* For generating the kubeconfig we generate a self signed CA, user key and cert and sign the user cert with the CA.
We also generate 4 signer CAs for the ingress and API and add thier certificates to the kubeconifg as the `CertificateAuthorityData`
 
* Recert requires 4 signer CAs keys for the ingress and API
* Recert also requires the CA cert that signed the client certificate


controllers/clusterconfig_controller.go
```diff
@@ -411,6 +413,26 @@ func (r *ClusterConfigReconciler) writeInputData(ctx context.Context, log logrus
 				}
 			}
 		}
+		CertificatesPath := filepath.Join(filesDir, certificatesDir)
```
carbonin, line 416: No need for this to be capitalized.

controllers/clusterconfig_controller.go
```diff
@@ -411,6 +413,26 @@ func (r *ClusterConfigReconciler) writeInputData(ctx context.Context, log logrus
...
+		certManager := certs.CertManager{CertificatesDir: CertificatesPath}
+		// TODO: handle user provided API and ingress certs
+		if err := certManager.GenerateAllCertificates(); err != nil {
+			return err
+		}
+		kubeconfigBytes, err := certManager.GenerateKubeConfig(config.Spec.Domain)
+		if err != nil {
+			return err
+		}
+		// TODO: change this to a secret
```
carbonin: I guess the goal would be to create a secret in the ACM hub containing this?

controllers/clusterconfig_controller.go
```diff
@@ -411,7 +413,44 @@ func (r *ClusterConfigReconciler) writeInputData(ctx context.Context, log logrus
...
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      config.Spec.ClusterName + "-admin-kubeconfig",
+				Namespace: config.Namespace,
+			},
+			Data: map[string][]byte{
+				"kubeconfig": kubeconfigBytes,
+			},
+			Type: corev1.SecretTypeOpaque,
+		}
+		err = r.Client.Create(ctx, kubeconfigSecret)
```
carbonin: Just make sure we have rbac for this.

internal/certs/certs_test.go
```diff
@@ -0,0 +1,118 @@
...
+			certificatePath: certPath,
+			keyPath:         keyPath,
+		}
+		ca, err := generateAndWriteCA(adminCA)
+		caCert := loadCACertFromFile(certPath)
+
+		userCert, _, err := generateAdminUserCertificate(ca)
+		Expect(err).NotTo(HaveOccurred())
+
+		// Verify that the client cert was signed by the given
```
carbonin, line 79: Was there supposed to be more to this comment? Given what?

carbonin (commented):
Few nits, looks good.

controllers/clusterconfig_controller.go
```diff
@@ -20,6 +20,8 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"github.com/openshift/cluster-relocation-service/internal/certs"
+	"k8s.io/apimachinery/pkg/api/errors"
```
hexfusion, line 24: minor: since we have logical seperation  already could we move these deps out of core.

controllers/clusterconfig_controller_test.go
```diff
@@ -311,6 +312,68 @@ var _ = Describe("Reconcile", func() {
 		Expect(err).To(HaveOccurred())
 	})
 
+	It("creates certificates", func() {
+
```
hexfusion, line 316: minor: drop newline?

internal/certs/certs.go
```diff
@@ -0,0 +1,153 @@
...
+	"crypto/x509/pkix"
+	"fmt"
+	"github.com/openshift/library-go/pkg/crypto"
+	"github.com/pkg/errors"
+	"k8s.io/apiserver/pkg/authentication/user"
+	"k8s.io/client-go/tools/clientcmd"
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
+	"path/filepath"
+	"time"
+)
```
hexfusion, line 13: minor: generally good style to breakout core/non core deps into separate blocks?

internal/certs/certs.go
```diff
@@ -0,0 +1,153 @@
...
+	user := user.DefaultInfo{Name: "system:admin"}
+	lifetime := time.Duration(crypto.DefaultCertificateLifetimeInDays) * 24 * time.Hour
+
+	cfg, err := ca.MakeClientCertificateForDuration(&user, lifetime)
+	if err != nil {
+		return nil, nil, errors.Wrap(err, "error making client certificate")
+	}
+	crt, key, err := cfg.GetPEMBytes()
+	if err != nil {
+		return nil, nil, errors.Wrap(err, "error getting PEM bytes for system:admin client certificate")
```
hexfusion, line 149: question: are we using errors.Wrap to collect stack trace? Otherwise
```suggestion
 		return nil, nil, fmt.Errorf("failed to get PEM bytes for system:admin client certificate: %w", err)
```

internal/certs/certs.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+
+	r.userCert, r.userKey, err = generateAdminUserCertificate(ca)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// Create signer keys and certificates and return a CA bundle from all CA certificates
```
omertuc, line 49:
```suggestion
// generateSignerCerts creates signer keys and certificates and return a CA bundle from all CA certificates
```

Doesn't follow go function documentation convention

internal/certs/certs.go
```diff
@@ -0,0 +1,153 @@
...
+	caBundle        []byte
+}
+type CertInfo struct {
+	keyPath         string
+	certificatePath string
+	commonName      string
+}
+
+func (r *CertManager) GenerateAllCertificates() error {
+	err := r.generateSignerCerts()
```
omertuc, line 28: I don't like the name signer certs, let's stick to the OCP convention of being verbose with certificate names as it's very easy to get lost in them - these are kube apiserver serving signing certs

internal/certs/certs.go
```diff
@@ -0,0 +1,153 @@
...
+			commonName: "kube-apiserver-localhost-signer",
+		},
+		{
+			keyPath:    filepath.Join(r.CertificatesDir, "service-network-serving-signer.key"),
+			commonName: "kube-apiserver-service-network-signer",
+		},
+		{
+			keyPath:    filepath.Join(r.CertificatesDir, "ingresskey-ingress-operator.key"),
+			commonName: fmt.Sprintf("%s@%d", "ingress-operator", time.Now().Unix()),
+		},
```
omertuc, line 68: I suggest we separate all the kube-apiserver certs from the ingress one, they're all bundled in the kubeconfig together but they're rather different logically

internal/certs/certs.go
```diff
@@ -0,0 +1,153 @@
...
+type CertInfo struct {
+	keyPath         string
+	certificatePath string
+	commonName      string
+}
+
+func (r *CertManager) GenerateAllCertificates() error {
+	err := r.generateSignerCerts()
+	if err != nil {
+		return err
```
omertuc, line 30: many unwrapped errors, it makes it hard to know where they come from

internal/certs/certs.go
```diff
@@ -0,0 +1,153 @@
...
+	if err != nil {
+		return nil, errors.Wrap(err, "error generating self signed CA")
+	}
+	return &crypto.CA{
+		SerialGenerator: &crypto.RandomSerialGenerator{},
+		Config:          newCAConfig,
+	}, nil
+}
+
+func generateAndWriteCA(certInfo CertInfo) (*crypto.CA, error) {
```
omertuc, line 127: You're using this function for both the client CA and the kapi/ingress serving CAs - does the openshift installer create those certificates in an identical manner? Or could there be differences between them that we missed?

internal/certs/certs.go
```diff
@@ -0,0 +1,153 @@
...
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
+	"path/filepath"
+	"time"
+)
+
+type CertManager struct {
+	CertificatesDir string
+	userCert        []byte
+	userKey         []byte
+	caBundle        []byte
```
omertuc, line 19:
```suggestion
	clusterCABundle        []byte
```

(to be more consistent with kubeconfig naming - the "ca bundle" goes under cluster, so it should be made clear in our naming

internal/certs/certs.go
```diff
@@ -0,0 +1,153 @@
...
+	"k8s.io/client-go/tools/clientcmd"
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
+	"path/filepath"
+	"time"
+)
+
+type CertManager struct {
+	CertificatesDir string
+	userCert        []byte
+	userKey         []byte
```
omertuc, line 18:
```suggestion
	userClientCert        []byte
	userClientKey         []byte
```

to make it clearer we're talking about `.users[].user.client*` things

internal/certs/certs.go
```diff
@@ -0,0 +1,153 @@
...
+	"github.com/openshift/library-go/pkg/crypto"
+	"github.com/pkg/errors"
+	"k8s.io/apiserver/pkg/authentication/user"
+	"k8s.io/client-go/tools/clientcmd"
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
+	"path/filepath"
+	"time"
+)
+
+type CertManager struct {
```
omertuc, line 15:
```suggestion
type KubeConfigCertManager struct {
```

To avoid giving the impression that this is some generic thing. This is focused solely on kubeconfig

internal/certs/certs.go
```diff
@@ -0,0 +1,153 @@
...
+	CertificatesDir string
```
omertuc, line 16: Can we outsource file handling to some external entity? Have this struct focus on the logical, pure side of things - create the keys and certs, create the kubeconfig, but not inside files. We're gonna need it when we move away from using files for everything. Also this is sensitive info, it should ideally avoid the disk as much as possible

omertuc (changes requested):
Some comments

internal/certs/certs_test.go
eranco74: Yes, I added the verification in line 83

internal/certs/certs.go
eranco74: This file contains a single line that actually writes certificates to a file.
We can change it once we no longer pass the certs as files.

internal/certs/certs.go
eranco74: I will need to verify that.

controllers/clusterconfig_controller.go
eranco74: We do

internal/certs/certs.go
eranco74: After checking the certs generated by the openshift installer I found the following differences:
1. All certs but the ingress get 10 Years validity (ingress is valid for 2 years).
2. All the certs have `x509.KeyUsageCertSign`
I'll update the code to generate matching certs.

eranco74: TBH seems to me that the client cert shouldn't have x509.ExtKeyUsageServerAuth.
There's no scenario in which this certificate holder might get used for a server.
The purpose of this cert is to be used by a client while making requests to the kube-apiserver for authentication.
So, instead of adding `x509.ExtKeyUsageServerAuth` here I'll send a PR to the installer for removing the serverAuth

eranco74:
/hold
This PR cause the imageserver to fail since it's generating files with names that are longer then the max allowed number of characters when creating diskfs 
https://github.com/diskfs/go-diskfs/blob/d67c42d83f8441f972e770521dd17aa732ee168a/filesystem/iso9660/directoryentry.go#L576
This code should be revisited once [MGMT-16483](https://issues.redhat.com/browse/MGMT-16483) is complete

osherdp:
/uncc
guess my review is not needed here

PR #23, opened 2024-01-04
Sync owners with openshift-kni/lifecycle-agent repo

omertuc:


OWNERS_ALIASES
```diff
@@ -2,11 +2,31 @@
...
     - oourfali
     - osherdp
-    - avishayt
-    - carbonin
+    - pixelsoccupied
+    - romfreiman
+    - sakhoury
+    - sudomakeinstall2
+    - tsorya
+    approvers:
```
carbonin, line 24: Should this approvers key be in here twice?

PR #24, opened 2024-01-08
MGMT-16502: Allow updating the clusterconfig status when BMH ref is set

eranco74:
When creating a ClusterConfig CR with BMH ref already set the controller will fail forever because the admission webhook will prevent it from updating the ClusterConfig status.
This PR fixes the above failure by allowing status updates regardless of the BMH ref.


api/v1alpha1/clusterconfig_webhook.go
```diff
@@ -55,13 +56,24 @@ func (r *ClusterConfig) ValidateUpdate(old runtime.Object) (admission.Warnings,
...
+		return nil, nil
+	}
 	if BMHRefsMatch(oldConfig.Spec.BareMetalHostRef, r.Spec.BareMetalHostRef) {
 		return nil, fmt.Errorf("Cannot update ClusterConfig when BareMetalHostRef is set, unset BareMetalHostRef before making changes")
 	}
-
 	return nil, nil
 }
 
+func IsStatusUpdate(oldConfig *ClusterConfig, r *ClusterConfig) bool {
```
carbonin, line 69: I don't think this needs to be exported.

PR #26, opened 2024-01-10
Use Hive ClusterInstall and ClusterDeployment as the main API

carbonin:
This PR moves the API from a custom type (`ClusterConfig`) to use `ImageClusterInstall` which conforms to the Hive `ClusterInstall` contract as well as a `ClusterDeployment`.

`ClusterDeployment` already contains some of the fields that were present in `ClusterConfig` so the controller now pulls the data from the referenced `ClusterDeployment`.

Both resources are now required for the reconciler to run.

This change was made to more closely align to other installation methods within MCE as well as to enable registration of the cluster with ACM (creation of a `ManagedCluster`) after the spoke SNO has been reconfigured.

Resolves https://issues.redhat.com/browse/MGMT-16432


controllers/imageclusterinstall_controller.go
```diff
@@ -44,64 +44,80 @@ import (
...
 		if err != nil {
 			log.Error(err)
 		}
 		return res, err
 	}
 
-	if res, err := r.writeInputData(ctx, log, config); !res.IsZero() || err != nil {
+	if clusterInstall.Spec.ClusterDeploymentRef == nil || clusterInstall.Spec.ClusterDeploymentRef.Name == "" {
+		log.Warn("ClusterDeploymentRef is unset, not reconciling")
+		return ctrl.Result{}, fmt.Errorf("missing ClusterDeploymentRef")
```
eranco74, line 104: Is it an error?
I think we can just return, once the CR gets updated with CDRef it will reconcile it again. 
Returning an error will spam the logs because it will get requeued

controllers/imageclusterinstall_controller.go
```diff
@@ -186,49 +202,72 @@ func (r *ClusterConfigReconciler) setHostConfiguredCondition(ctx context.Context
 		cond.Message = err.Error()
 	}
 
-	patch := client.MergeFrom(config.DeepCopy())
-	meta.SetStatusCondition(&config.Status.Conditions, cond)
-	return r.Status().Patch(ctx, config, patch)
+	patch := client.MergeFrom(clusterInstall.DeepCopy())
+	meta.SetStatusCondition(&clusterInstall.Status.ConfigConditions, cond)
```
eranco74, line 206: I see that nothing really changed, but can you please explain what's going on here?
What's the patch for?

controllers/imageclusterinstall_controller.go
```diff
@@ -186,49 +202,72 @@ func (r *ClusterConfigReconciler) setHostConfiguredCondition(ctx context.Context
...
 	bmh := &bmh_v1alpha1.BareMetalHost{}
 	bmhName := obj.GetName()
 	bmhNamespace := obj.GetNamespace()
 
 	if err := r.Get(ctx, types.NamespacedName{Name: bmhName, Namespace: bmhNamespace}, bmh); err != nil {
 		return []reconcile.Request{}
 	}
-	ccList := &relocationv1alpha1.ClusterConfigList{}
-	if err := r.List(ctx, ccList); err != nil {
+	iciList := &relocationv1alpha1.ImageClusterInstallList{}
```
eranco74, line 218: Not for this PR, but I think we want some annotation on the BMH that say the BMH is pre-installed with template SNO

controllers/imageclusterinstall_controller.go
```diff
@@ -186,49 +202,72 @@ func (r *ClusterConfigReconciler) setHostConfiguredCondition(ctx context.Context
...
+	cdName := obj.GetName()
+	cdNamespace := obj.GetNamespace()
+
+	cd := &hivev1.ClusterDeployment{}
+	if err := r.Get(ctx, types.NamespacedName{Name: cdName, Namespace: cdNamespace}, cd); err != nil {
+		return []reconcile.Request{}
+	}
+
+	if cd.Spec.ClusterInstallRef != nil &&
+		cd.Spec.ClusterInstallRef.Group == relocationv1alpha1.Group &&
```
eranco74, line 257: `relocationv1alpha1`?

controllers/imageclusterinstall_controller.go
```diff
@@ -325,8 +365,8 @@ func (r *ClusterConfigReconciler) removeBMHImage(ctx context.Context, bmhRef *re
 	return nil
 }
 
-func (r *ClusterConfigReconciler) configDirs(config *relocationv1alpha1.ClusterConfig) (string, string, error) {
-	lockDir := filepath.Join(r.Options.DataDir, "namespaces", config.Namespace, config.Name)
+func (r *ImageClusterInstallReconciler) configDirs(clusterInstall *relocationv1alpha1.ImageClusterInstall) (string, string, error) {
```
eranco74, line 368: In some functions, it's `ici` in others it's `clusterInstall`.
Try sticking with one of them (I prefer `ibi`)

controllers/imageclusterinstall_controller.go
```diff
@@ -422,16 +462,27 @@ func (r *ClusterConfigReconciler) writeInputData(ctx context.Context, log logrus
...
 		}
 		return ctrl.Result{RequeueAfter: time.Second * 5}, nil
 	}
 
 	return ctrl.Result{}, nil
 }
 
-func (r *ClusterConfigReconciler) writeClusterInfo(info *clusterinfo.ClusterInfo, file string) error {
+func (r *ImageClusterInstallReconciler) writeClusterInfo(ici *relocationv1alpha1.ImageClusterInstall, cd *hivev1.ClusterDeployment, file string) error {
+	info := clusterinfo.ClusterInfo{
```
eranco74, line 475: :+1:

api/v1alpha1/imageclusterinstall_types.go
```diff
@@ -54,17 +73,25 @@ type ClusterConfigSpec struct {
 
 	// ExtraManifestsRefs is list of config map references containing additional manifests to be applied to the relocated cluster.
 	// +optional
+	// TODO: use ClusterDeployment.Spec.Provisioning.ManifestsConfigMapRef/ManifestsSecretRef?
```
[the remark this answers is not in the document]
carbonin: :+1: right, I'll leave it here then

api/v1alpha1/imageclusterinstall_types.go
```diff
@@ -34,14 +34,33 @@ const (
 	HostConfigurationSucceededMessage = "Configuration image is attached to the referenced host"
 )
 
-// ClusterConfigSpec defines the desired state of ClusterConfig
-type ClusterConfigSpec struct {
-	clusterinfo.ClusterInfo `json:",inline"`
+// ImageClusterInstallSpec defines the desired state of ImageClusterInstall
+type ImageClusterInstallSpec struct {
```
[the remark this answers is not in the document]
carbonin: That hasn't been implemented yet in this operator as far as I know

controllers/imageclusterinstall_controller.go
carbonin: Good point. I'll log an error but not return one

controllers/imageclusterinstall_controller.go
carbonin: This is a more targeted way to update a resource. It is better at avoiding the resource version errors.

controllers/imageclusterinstall_controller.go
carbonin: That's the group name for this operator.

I figure we can change it once we decide on the new repo name.

api/v1alpha1/imageclusterinstall_types.go
```diff
@@ -34,14 +34,33 @@ const (
...
+	// Version is the target OCP version for the cluster
+	// TODO: should this use ImageSetRef?
+	Version string `json:"version,omitempty"`
 
-	// PullSecretRef is a reference to new cluster-wide pull secret.
-	// If defined, it will replace the secret located at openshift-config/pull-secret.
-	// The type of the secret must be kubernetes.io/dockerconfigjson.
-	PullSecretRef *corev1.LocalObjectReference `json:"pullSecretRef,omitempty"`
+	// MasterIP is the desired IP for the host
+	MasterIP string `json:"masterIP,omitempty"`
```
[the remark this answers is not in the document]
tsorya: We use it for ip change operation. I assume we need to change logic in ibi but in ibu it is pretty natural to take ip from the running cluster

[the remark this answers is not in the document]
tsorya: Another option will be to start nodeip service manually and take ip it had chosen. We can do it but for IBU it was over-complication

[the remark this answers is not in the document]
carbonin: Also should we not use "master" in the public API even though that's what we're using in the cluster info?
IIRC there was some push to move away from the word generally. Maybe "HostIP" would be enough?

Also might be worth thinking about what these would look like if we do eventually get multi-node working for this. We probably wouldn't want a single host IP in the cluster install in that case.

controllers/imageclusterinstall_controller.go
```diff
@@ -456,7 +471,18 @@ func (r *ImageClusterInstallReconciler) writeInputData(ctx context.Context, log
...
+		Version:         ici.Spec.Version,
+		Domain:          cd.Spec.BaseDomain,
+		ClusterName:     cd.Spec.ClusterName,
+		MasterIP:        ici.Spec.MasterIP,
+		ReleaseRegistry: ici.Spec.ReleaseRegistry,
+		Hostname:        ici.Spec.Hostname,
+	}
+	if ici.Spec.ClusterMetadata != nil {
+		info.ClusterID = ici.Spec.ClusterMetadata.ClusterID
```
carbonin, line 485: Using `ClusterMetadata` here is a bit of a problem.

Testing this is giving me this error when trying to create the ICI:
```
The ImageClusterInstall "test-sno" is invalid: 
* spec.clusterMetadata.adminKubeconfigSecretRef: Required value
* spec.clusterMetadata.infraID: Required value
```

If we're requiring the user to provide the cluster ID here they're definitely not going to have values for these other two (and I'm not even sure what we'd use for `infraID`)

config/crd/bases/relocation.openshift.io_imageclusterinstalls.yaml
```diff
@@ -188,20 +184,13 @@ spec:
...
-                type: object
-                x-kubernetes-map-type: atomic
-              release_registry:
+              releaseRegistry:
+                description: 'ReleaseRegistry is the registry of the installed release
+                  image TODO: should this also use the registry of ImageSetRef?'
                 type: string
               version:
+                description: 'Version is the target OCP version for the cluster TODO:
+                  should this use ImageSetRef?'
```
carbonin, line 193: Related to this I'm getting an error trying to create the `ImageClusterInstall` in testing saying that this is required.

I don't think we can change that and still conform to the hive `ClusterInstall` API

api/v1alpha1/imageclusterinstall_types.go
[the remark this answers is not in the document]
carbonin: What was the decision on this can we do without it @tsorya ?

api/v1alpha1/imageclusterinstall_types.go
```diff
@@ -34,14 +34,36 @@ const (
...
+	ImageSetRef hivev1.ClusterImageSetReference `json:"imageSetRef"`
+
+	// ClusterMetadata contains metadata information about the installed cluster.
+	// This must be set as soon as all the information is available.
+	// +optional
+	ClusterMetadata *hivev1.ClusterMetadata `json:"clusterMetadata"`
+
+	// Version is the target OCP version for the cluster
+	// TODO: should this use ImageSetRef?
+	Version string `json:"version,omitempty"`
```
carbonin, line 53: This should use ImageSetRef

api/v1alpha1/imageclusterinstall_types.go
```diff
@@ -34,14 +34,36 @@ const (
...
-	// PullSecretRef is a reference to new cluster-wide pull secret.
-	// If defined, it will replace the secret located at openshift-config/pull-secret.
-	// The type of the secret must be kubernetes.io/dockerconfigjson.
-	PullSecretRef *corev1.LocalObjectReference `json:"pullSecretRef,omitempty"`
+	// Hostname is the desired hostname for the host
+	Hostname string `json:"hostname,omitempty"`
+
+	// ReleaseRegistry is the registry of the installed release image
+	// TODO: should this also use the registry of ImageSetRef?
+	ReleaseRegistry string `json:"releaseRegistry,omitempty"`
```
carbonin, line 63: @omertuc what exactly is this? Would it have anything to do with the release image?

api/v1alpha1/imageclusterinstall_types.go
carbonin, line 53: See what it takes to set this using cluster image set.

api/v1alpha1/imageclusterinstall_types.go
```diff
@@ -34,14 +34,36 @@ const (
...
+
+	// ClusterID is the desired ID for the new cluster
+	ClusterID string `json:"clusterID,omitempty"`
```
carbonin, line 66: Maybe make a mutating webhook to generate this if it wasn't provided.

api/v1alpha1/imageclusterinstall_types.go
```diff
@@ -61,10 +83,17 @@ type ClusterConfigSpec struct {
...
+	// Conditions is a list of conditions associated with syncing to the cluster.
+	// +optional
+	Conditions []hivev1.ClusterInstallCondition `json:"conditions,omitempty"`
+
+	// InstallRestarts is the total count of container restarts on the clusters install job.
+	InstallRestarts int `json:"installRestarts,omitempty"`
+
 	BareMetalHostRef *BareMetalHostReference `json:"bareMetalHostRef,omitempty"`
-	Conditions       []metav1.Condition      `json:"conditions,omitempty"`
+	ConfigConditions []metav1.Condition      `json:"configConditions,omitempty"`
```
carbonin, line 96: @eranco74 any thoughts on the name here?

carbonin: Or if we should force these into the existing `Conditions` field.

config/crd/bases/relocation.openshift.io_imageclusterinstalls.yaml
carbonin: Moving this conversation to the types.go file.

controllers/imageclusterinstall_controller.go
eranco74: Cool, thanks

api/v1alpha1/imageclusterinstall_types.go
omertuc: LCA will do the generating if you leave this empty anyway, so you don't have to bother making a mutating webhook

controllers/imageclusterinstall_controller.go
eranco74: Let's change it once the repo name is updated

api/v1alpha1/imageclusterinstall_types.go
omertuc: [Baffled](https://github.com/omertuc/lifecycle-agent/commit/e3a82df9d37f6f5e8c08172c8f0246c470c35b0c#diff-e6ccbfc77f9d37a80162c13f8b94d2044699b8d090451a8848d2b4686b6ff815R45) me as well

controllers/imageclusterinstall_controller.go
eranco74: Let's not use clusterMetadata

eranco74:
This is a great step towards the API. 
/approve

api/v1alpha1/imageclusterinstall_types.go
eranco74: @donpenney , @tsorya does any of you know what this field is and what it's for?

donpenney: This is specifically for the release image reference, which I think is explicitly the mirror. It required special handling in the case of disconnected registries as a result, as it uses the explicit registry reference and not the original quay.io ref like other images. @tsorya did a lot of work to handle these cases, and could likely provide a better explanation

api/v1alpha1/imageclusterinstall_types.go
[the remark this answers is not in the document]
tsorya: Adding code that will handle empty param case

controllers/imageclusterinstall_controller.go
carbonin: Do you want to change this in this PR now that the name is changed or do it in a follow up? My vote is a follow up FWIW

api/v1alpha1/imageclusterinstall_types.go
carbonin: Removed this.

controllers/imageclusterinstall_controller.go
eranco74: In a followup

api/v1alpha1/imageclusterinstall_types.go
eranco74: So from what I understand, this is required in case the seed image was installed using a release image from a mirror registry that is different from the registry the cluster should use.
We use this field to override the opensfhit-cluster-version deployment `release-image` parameter with an updated release image FQDN that uses the registry the cluster should use. 

Note: this is relevant for cases where the release image is specified with the FQDN from the mirror registry and not redirected via ICSP or IDSP.

api/v1alpha1/imageclusterinstall_types.go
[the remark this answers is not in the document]
eranco74: After the talk we had Yesterday it seems that we can expect to get this information from the user.
So let's keep it as optional and rename it to NodeIP.
In any case @tsorya we should handle empty string, do you have a ticket for it?

controllers/imageclusterinstall_controller_test.go
```diff
@@ -107,11 +122,10 @@ var _ = Describe("Reconcile", func() {
 			Domain:          "example.com",
 			ClusterName:     "thingcluster",
 			MasterIP:        "192.0.2.1",
-			ReleaseRegistry: "registry.example.com",
+			ReleaseRegistry: "registry.example.com/releases",
```
eranco74, line 125: LCA get the reigstry [this way](https://github.com/openshift-kni/lifecycle-agent/blob/9ff71276db124e563af5adad60f35f53eb15241a/utils/client_helper.go#L190), meaning in LCA will not expect ot the `/release` in the registry name

api/v1alpha1/imageclusterinstall_types.go
eranco74: Remind me why we don't want to get it as input from the user

controllers/imageclusterinstall_controller_test.go
carbonin: Okay so it only wants the registry hostname?

carbonin: That would make this a lot less error prone actually

carbonin: Done

api/v1alpha1/imageclusterinstall_types.go
carbonin: Because it is pretty much duplicated information, right?
It should be part of the cluster image set and the true source of this is the seed image which is already installed.

api/v1alpha1/imageclusterinstall_types.go
carbonin: The only potential issue I saw with that is the way the hive `ClusterInstallCondition` was written.
It has specific types and I'm wondering if anything would have a problem with us not using one of the defined ones because I don't think our conditions fit in the existing ones.

I can give this a try though.

ref: https://github.com/openshift/hive/blob/de169ddf445c3e064cfa8773de5af8ec00f649be/apis/hive/v1/clusterinstall_conditions.go

api/v1alpha1/imageclusterinstall_types.go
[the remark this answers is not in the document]
carbonin: Unresolving this as I still need to change the name.

eranco74:
This should get the prow jobs running
https://github.com/openshift/release/pull/47802

PR #27, opened 2024-01-19
MGMT-16618: Make hive APIs work

carbonin:
This PR makes a few changes to ensure the new hive APIs are usable.

It adds the hive contract label to the CRD which is required for hive to recognize this type as a valid
ClusterInstall sub-type. Without this the hive admission webhook returns
an error like:

```
Error from server (Invalid): error when creating "cluster.yaml": admission webhook "clusterdeploymentvalidators.admission.hive.openshift.io" denied the request: ClusterDeployment.hive.openshift.io "test-sno" is invalid: spec.clusterInstallRef.kind: Unsupported value: "ImageClusterInstall": supported values: "extensions.hive.openshift.io/v1beta1, Kind=AgentClusterInstall", "hiveinternal.openshift.io/v1alpha1, Kind=FakeClusterInstall"
```

It also changes the API group to extensions.hive.openshift.io. This is required so that hive has permissions to access and edit the new
ClusterInstall type.

See
https://github.com/openshift/hive/blob/de169ddf445c3e064cfa8773de5af8ec00f649be/docs/enhancements/cluster-install-apis.md#external-clusterinstall-rbac
for details

Lastly it adds clusterimageset rbac which was missed in the previous PR.

Resolves https://issues.redhat.com/browse/MGMT-16618

### DIFF
--- a/config/crd/patches/hive_contract_label.yaml
+++ b/config/crd/patches/hive_contract_label.yaml
@@ -3,4 +3,4 @@ kind: CustomResourceDefinition
 metadata:
   labels:
     contracts.hive.openshift.io/clusterinstall: "true"
-  name: imageclusterinstalls.relocation.openshift.io
+  name: imageclusterinstalls.extensions.hive.openshift.io
--- a/config/crd/patches/hive_contract_label.yaml
+++ b/config/crd/patches/hive_contract_label.yaml
@@ -1,0 +1,6 @@
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  labels:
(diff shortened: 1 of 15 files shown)


config/crd/patches/hive_contract_label.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  labels:
+    contracts.hive.openshift.io/clusterinstall: "true"
```
eranco74, line 5: :+1:

eranco74 (commented):
/lgtm

PR #28, opened 2024-01-19
MGMT-16554: Only fail the admission webhook if the spec was changed

carbonin:
Previously this was failing for status, and is now failing for metadata (specifically when removing the finalizer).

Resolves https://issues.redhat.com/browse/MGMT-16554


eranco74 (commented):
/lgtm
/approve

PR #29, opened 2024-01-19
MGMT-16614: Update names referencing relocation

carbonin:
This commit changes any names referencing relocation to something that agrees with the new repo name. This includes the module/import path as well as the operator namespace and all the object names in the operator deploy manifests.

This doesn't handle the group name as that is involved with the hive API work and is solved in https://github.com/openshift/image-based-install-operator/pull/27

Resolves https://issues.redhat.com/browse/MGMT-16614


carbonin:
Marking this as WIP until I can test it on top of the hive API changes to ensure it deploys in the new namespace.

eranco74 (commented):
/lgtm
Should we update the image name in openshift/release as well?

PR #33, opened 2024-01-24
MGMT-16650 Regenerate cluster crypto only when needed

eranco74:
Prior to this change, IBIO renders the cluster crypto 4 times in case of happy flow. It's useless time consuming and error prone (we might accidintly override the cluster kubeconfig) This change will generate new crypto in case the current cluster crypto doesn't match the current cluster name and basedomain.

### DIFF
--- a/controllers/imageclusterinstall_controller.go
+++ b/controllers/imageclusterinstall_controller.go
@@ -79,6 +79,7 @@ const (
 	extraManifestsDir           = "extra-manifests"
 	manifestsDir                = "manifests"
 	networkConfigDir            = "network-configuration"
+	kubeconfig                  = "kubeconfig"
 	clusterInstallFinalizerName = "imageclusterinstall." + v1alpha1.Group + "/deprovision"
 	caBundleFileName            = "tls-ca-bundle.pem"
 )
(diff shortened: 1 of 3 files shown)


controllers/imageclusterinstall_controller.go
```diff
@@ -485,19 +488,35 @@ func (r *ImageClusterInstallReconciler) writeInputData(ctx context.Context, log
...
+	data, err := os.ReadFile(clusterInfoFilePath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to write cluster info: %w", err)
+	}
+	clusterInfo := lca_api.SeedReconfiguration{}
+	err = json.Unmarshal(data, &clusterInfo)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal cluster info: %w", err)
+	}
+	if clusterInfo.ClusterName != cd.Spec.ClusterName || clusterInfo.BaseDomain != cd.Spec.BaseDomain {
```
omertuc, line 501: Why these two parameters in particular are used to guard from redundant regeneration? (preferably answer as a code comment)

Also it's not obvious that the goal of this if condition here is to trigger a regeneration, so maybe clarify that

controllers/imageclusterinstall_controller.go
```diff
@@ -458,12 +459,23 @@ func (r *ImageClusterInstallReconciler) writeInputData(ctx context.Context, log
 				}
 			}
 		}
-
-		crypto, err := r.generateClusterCrypto(cd, ctx)
-		if err != nil {
-			return err
+		clusterInfoFilePath := filepath.Join(clusterConfigPath, "manifest.json")
+		clusterInfo, err := r.getClusterInfoFromFile(clusterInfoFilePath)
```
carbonin, line 463: Do we check or care about err here?

controllers/imageclusterinstall_controller_test.go
```diff
@@ -154,6 +154,88 @@ var _ = Describe("Reconcile", func() {
 		Expect(*infoOut).To(Equal(info))
 	})
 
+	It("keep cluster crypto if name and base domain didnt' change", func() {
```
carbonin, line 157: `didn't`

controllers/imageclusterinstall_controller.go
```diff
@@ -458,12 +459,23 @@ func (r *ImageClusterInstallReconciler) writeInputData(ctx context.Context, log
...
+			// handle first cert generation
+			crypto, err = r.generateClusterCrypto(ctx, cd)
+			if err != nil {
+				return err
+			}
+		} else {
+			crypto = &clusterInfo.KubeconfigCryptoRetention
+			if clusterInfo.ClusterName != cd.Spec.ClusterName || clusterInfo.BaseDomain != cd.Spec.BaseDomain {
+				// in case the cluster name or baseDomain changed we need a new kubeconfig, so we just regenerate all crypto
+				crypto, err = r.generateClusterCrypto(ctx, cd)
```
carbonin, line 475: This err should be checked too.

carbonin (approved):

controllers/imageclusterinstall_controller.go
eranco74: After discussing it f2f we agreed that the optimization isn't worth the complexity and the additional edge cases we will need to handle.
So in case the cluster name or based domain changed we should just regenerate all cluster crypto + kubeconfig

PR #35, opened 2024-02-01
MGMT-16609 Add SSHPublicKey config option

eranco74:


api/v1alpha1/imageclusterinstall_types.go
```diff
@@ -59,6 +59,11 @@ type ImageClusterInstallSpec struct {
 	// Hostname is the desired hostname for the host
 	Hostname string `json:"hostname,omitempty"`
 
+	// SSHKey is the public Secure Shell (SSH) key to provide access to
+	// instances. Equivalent to install-config.yaml's sshKey.
+	// This key will be added to the host to allow ssh access
+	SSHKey string `json:"ssh_key,omitempty"`
```
carbonin, line 65: Would this be better as `SSHPublicKey`? Or is matching install config's API more of a priority?

eranco74: I don't mind.

carbonin (approved):

PR #38, opened 2024-02-07
MGMT-16632: Use hive conditions on ImageClusterInstall

carbonin:
Instead of maintaining a separate list of conditions use the ones hive is watching for as we'll need to set those properly anyway.

This PR also sets the required conditions for an installed cluster after configuring the BMH with the image.

Checking if the cluster actually installed properly is left as a follow-up.

Resolves https://issues.redhat.com/browse/MGMT-16632


bundle/manifests/extensions.hive.openshift.io_imageclusterinstalls.yaml
```diff
@@ -185,6 +185,11 @@ spec:
               nodeIP:
                 description: NodeIP is the desired IP for the host
                 type: string
+              ssh_key:
```
eranco74, line 188: Underscore?

controllers/conditions.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+func setClusterInstallCondition(conditions *[]hivev1.ClusterInstallCondition, newCondition hivev1.ClusterInstallCondition) {
+	if conditions == nil {
+		return
+	}
+	now := metav1.NewTime(time.Now())
+	existingCondition := findCondition(*conditions, newCondition.Type)
+	if existingCondition == nil {
+		if newCondition.LastTransitionTime.IsZero() {
```
eranco74, line 30: Curious, Who is setting the `LastProbeTime` and `LastTransitionTime` on the newCondition?
From the code it seems that they are never set except for here so why bother checking for IsZero() ?

controllers/conditions.go
```diff
@@ -0,0 +1,133 @@
...
+		cond.Reason = v1alpha1.ImageNotReadyReason
+		cond.Message = err.Error()
+	}
+
+	patch := client.MergeFrom(ici.DeepCopy())
+	setClusterInstallCondition(&ici.Status.Conditions, cond)
+	return r.Status().Patch(ctx, ici, patch)
+}
+
+func (r *ImageClusterInstallReconciler) setHostConfiguredCondition(ctx context.Context, ici *v1alpha1.ImageClusterInstall, err error) error {
```
eranco74, line 98: Where do you call this?

controllers/imageclusterinstall_controller.go
```diff
@@ -190,47 +190,14 @@ func (r *ImageClusterInstallReconciler) Reconcile(ctx context.Context, req ctrl.
...
-		Status:  metav1.ConditionTrue,
-		Reason:  v1alpha1.HostConfiguraionSucceededReason,
-		Message: v1alpha1.HostConfigurationSucceededMessage,
-	}
 
-	if err != nil {
-		cond.Status = metav1.ConditionFalse
-		cond.Reason = v1alpha1.HostConfiguraionFailedReason
-		cond.Message = err.Error()
+		if err := r.setClusterInstalledConditions(ctx, ici); err != nil {
```
eranco74, line 194: Seems premature but I guess for the time being we can't really tell so why not

bundle/manifests/extensions.hive.openshift.io_imageclusterinstalls.yaml
carbonin: Ah yeah, the bundle wasn't generated with your PR and you set it with the underscore, but I can fix that separately.

ref: https://github.com/openshift/image-based-install-operator/pull/35/files#diff-126f7ef4b1c14a1e0e4386760e8a5398759328cc3fe36a104ee0cafbc6fa932dR65

controllers/conditions.go
carbonin: Ah this is just a copy of the core function that does the same thing, but with core conditions.

I can trim this down to only what we need.

controllers/conditions.go
carbonin: The function already existed it just set a different condition before.

Currently it's called here https://github.com/openshift/image-based-install-operator/blob/6f767d683a7091a4831fe27b8b034a79d4f54e8e/controllers/imageclusterinstall_controller.go#L167-L172

carbonin: I debated changing the name to match the new condition type, but these functions are also setting the reason and message to the IBIO-specific ones so I just left it since we'd want it called at the same time.

controllers/imageclusterinstall_controller.go
carbonin: I'm actually also already setting the CD to installed as well so I figure we can change both later if it becomes a problem and we want to implement some kind of monitoring

ref: https://github.com/openshift/image-based-install-operator/blob/6f767d683a7091a4831fe27b8b034a79d4f54e8e/controllers/imageclusterinstall_controller.go#L784

controllers/conditions.go
carbonin: This is the function I was referencing FWIW https://github.com/kubernetes/apimachinery/blob/v0.28.3/pkg/api/meta/conditions.go#L30

bundle/manifests/extensions.hive.openshift.io_imageclusterinstalls.yaml
eranco74: Thanks

PR #39, opened 2024-02-08
MGMT-16714 Create kubeadmin password

eranco74:
Create a password for the kubeadmin user and store it in a secret
Set  the ClusterDeployment in the secret owner ref
Add the password hash to the SeedReconfiguration

### DIFF
--- a/controllers/imageclusterinstall_controller.go
+++ b/controllers/imageclusterinstall_controller.go
@@ -618,7 +618,7 @@ func (r *ImageClusterInstallReconciler) writeImageDigestSourceToFile(imageDigest
 
 	imageDigestMirrorSet := &apicfgv1.ImageDigestMirrorSet{
 		TypeMeta: metav1.TypeMeta{
-			APIVersion: apicfgv1.SchemeGroupVersion.String(),
+			APIVersion: apicfgv1.GroupVersion.String(),
 			Kind:       "ImageDigestMirrorSet",
 		},
 		ObjectMeta: metav1.ObjectMeta{
--- a/controllers/imageclusterinstall_controller.go
+++ b/controllers/imageclusterinstall_controller.go
@@ -57,6 +57,7 @@ import (
 	hivev1 "github.com/openshift/hive/apis/hive/v1"
 	"github.com/openshift/image-based-install-operator/api/v1alpha1"
 	"github.com/openshift/image-based-install-operator/internal/certs"
+	"github.com/openshift/image-based-install-operator/internal/credentials"
 	"github.com/openshift/image-based-install-operator/internal/filelock"
 	"github.com/sirupsen/logrus"
 )
@@ -72,6 +73,7 @@ type ImageClusterInstallReconcilerOptions struct {
 // ImageClusterInstallReconciler reconciles a ImageClusterInstall object
 type ImageClusterInstallReconciler struct {
 	client.Client
+	credentials.Credentials
 	Log         logrus.FieldLogger
 	Scheme      *runtime.Scheme
 	Options     *ImageClusterInstallReconcilerOptions
@@ -89,7 +91,6 @@ const (
 	extraManifestsDir            = "extra-manifests"
 	manifestsDir                 = "manifests"
 	nmstateCMKey                 = "network-config"
-	kubeconfig                   = "kubeconfig"
 	clusterInstallFinalizerName  = "imageclusterinstall." + v1alpha1.Group + "/deprovision"
 	caBundleFileName             = "tls-ca-bundle.pem"
 	imageDigestMirrorSetFileName = "image-digest-sources.json"
@@ -190,7 +191,7 @@ func (r *ImageClusterInstallReconciler) Reconcile(ctx context.Context, req ctrl.
 			return ctrl.Result{}, err
 		}
 
-		if err := r.setClusterInstallMetadata(ctx, ici, kubeconfigSecretName(clusterDeployment)); err != nil {
+		if err := r.setClusterInstallMetadata(ctx, ici, clusterDeployment.Name); err != nil {
 			log.WithError(err).Error("failed to set ImageClusterInstall metadata")
 			return ctrl.Result{}, err
 		}
@@ -440,7 +441,11 @@ func (r *ImageClusterInstallReconciler) writeInputData(ctx context.Context, log 
 		} else {
 			crypto = &clusterInfo.KubeconfigCryptoRetention
 		}
-		if err := r.writeClusterInfo(ctx, log, ici, cd, *crypto, psData, clusterInfoFilePath, clusterInfo); err != nil {
+		kubeadminPasswordHash, err := r.Credentials.EnsureAdminPasswordSecret(ctx, cd)
+		if err != nil {
+			return fmt.Errorf("failed to ensure admin password secret: %w", err)
+		}
+		if err := r.writeClusterInfo(ctx, log, ici, cd, *crypto, psData, kubeadminPasswordHash, clusterInfoFilePath, clusterInfo); err != nil {
 			return fmt.Errorf("failed to write cluster info: %w", err)
 		}
 		return nil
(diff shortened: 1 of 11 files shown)


controllers/imageclusterinstall_controller.go
```diff
@@ -709,7 +684,7 @@ func (r *ImageClusterInstallReconciler) writeImageDigestSourceToFile(imageDigest
 
 	imageDigestMirrorSet := &apicfgv1.ImageDigestMirrorSet{
 		TypeMeta: metav1.TypeMeta{
-			APIVersion: apicfgv1.SchemeGroupVersion.String(),
+			APIVersion: apicfgv1.GroupVersion.String(),
```
carbonin, line 687: What does this change do? I imagine it's not related to the kubeadmin password, right?

controllers/imageclusterinstall_controller.go
```diff
@@ -742,10 +717,14 @@ func (r *ImageClusterInstallReconciler) setClusterInstallMetadata(ctx context.Co
...
 	}
 
+	kubeconfigSecret := credentials.KubeconfigSecretName(clusterDDeploymentName)
+	kubeadminPasswordSecret := credentials.KubeadminPasswordSecretName(clusterDDeploymentName)
 	if ici.Spec.ClusterMetadata != nil &&
 		ici.Spec.ClusterMetadata.ClusterID == clusterInfo.ClusterID &&
 		ici.Spec.ClusterMetadata.InfraID == clusterInfo.InfraID &&
-		ici.Spec.ClusterMetadata.AdminKubeconfigSecretRef.Name == kubeconfigSecret {
+		ici.Spec.ClusterMetadata.AdminKubeconfigSecretRef.Name == kubeconfigSecret &&
+		ici.Spec.ClusterMetadata.AdminPasswordSecretRef.Name == kubeadminPasswordSecret {
```
carbonin, line 726: I think this needs to be set below (line 737-ish) as you did in `setClusterDeploymentMetadata`

internal/credentials/credentialls.go
```diff
@@ -0,0 +1,176 @@
...
+)
+
+type Credentials struct {
+	client.Client
+	Log    logrus.FieldLogger
+	Scheme *runtime.Scheme
+}
+
+func (r *Credentials) EnsureKubeconfigSecret(ctx context.Context, cd *hivev1.ClusterDeployment, kubeconfigBytes []byte) error {
+	//		Type: corev1.SecretTypeOpaque,
```
carbonin, line 37: Remove this?

internal/credentials/credentialls.go
```diff
@@ -0,0 +1,176 @@
...
+
+func KubeadminPasswordSecretName(clusterDeploymentName string) string {
+	return clusterDeploymentName + "-admin-password"
+}
+
+func generateKubeadminPassword() (string, error) {
+	const (
+		lowerLetters = "abcdefghijkmnopqrstuvwxyz"
+		upperLetters = "ABCDEFGHIJKLMNPQRSTUVWXYZ"
+		digits       = "23456789"
```
carbonin, line 135: Missing 1 and 0?
Was that intentional?

internal/credentials/credentialls.go
```diff
@@ -0,0 +1,176 @@
...
+			j := n.Int64()
+			password = password[0:j] + newchar + password[j:]
+		}
+	}
+	pw := []rune(password)
+	for _, replace := range []int{5, 11, 17} {
+		pw[replace] = '-'
+	}
+	return string(pw), nil
+	//	bytes, err := bcrypt.GenerateFromPassword([]byte(password), bcrypt.DefaultCost)
```
carbonin, line 163: Remove this?

internal/credentials/credentialls.go
```diff
@@ -0,0 +1,176 @@
...
+		Name:               cd.Name,
+		UID:                cd.UID,
+		BlockOwnerDeletion: pointer.Bool(true),
+	}}
+	mutateFn := func() error {
+		// Update the Secret object with the desired data
+		secret.Data = data
+		return nil
+	}
+	op, err := controllerutil.CreateOrUpdate(ctx, r.Client, secret, mutateFn)
```
carbonin, line 115: I feel like the controller should be the one interacting with the API objects.

This package should probably just be creating the credential values similar to the separation we have in the certs package.

carbonin (changes requested):

controllers/imageclusterinstall_controller.go
eranco74: True, it's just a nit, SchemeGroupVersion was deprecated in favor of GroupVersion.
I'll move it to a separate commit

internal/credentials/credentialls.go
eranco74: Copied from the installer, looked at the same code in hypershift no 1 and 0 there as well.
:man_shrugging:

internal/credentials/credentialls.go
eranco74: I wanted to move everything related to the cluster credentials from the controller to this package (including the API objects).
I even considered moving the method that generates the password to another package (utils.go?)
Let's discuss it on Monday.

internal/credentials/credentialls.go
```diff
@@ -0,0 +1,176 @@
...
+			if err != nil {
+				return "", fmt.Errorf("failed to generate password hash: %w", err)
+			}
+			return string(passwordHash), nil
+		}
+	} else if !errors.IsNotFound(err) {
+		return "", fmt.Errorf("failed to get kubeadmin password secret: %w", err)
+	}
+	r.Log.Infof("Generating admin password")
+	password, err := generateKubeadminPassword()
```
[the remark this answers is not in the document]
eranco74: Agree, I'll update `EnsureKubeconfigSecret` to work the same.
Mind if it's in a separate PR? it will take some refactoring that will make the PR more complicated than it should be.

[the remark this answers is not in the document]
eranco74: NM, updated this PR with the change

controllers/imageclusterinstall_controller.go
```diff
@@ -451,24 +452,13 @@ func (r *ImageClusterInstallReconciler) writeInputData(ctx context.Context, log
...
-		} else {
-			crypto = &clusterInfo.KubeconfigCryptoRetention
-			if clusterInfo.ClusterName != cd.Spec.ClusterName || clusterInfo.BaseDomain != cd.Spec.BaseDomain {
-				// in case the cluster name or baseDomain changed we need a new kubeconfig, so we just regenerate all crypto
-				crypto, err = r.generateClusterCrypto(ctx, cd)
-				if err != nil {
-					return err
-				}
-			}
+		crypto, err := r.EnsureKubeconfigSecret(ctx, cd, clusterInfo)
```
carbonin, line 455: Check this error

controllers/imageclusterinstall_controller.go
```diff
@@ -451,24 +452,13 @@ func (r *ImageClusterInstallReconciler) writeInputData(ctx context.Context, log
...
+
+		kubeadminPasswordHash, err := r.Credentials.EnsureAdminPasswordSecret(ctx, cd)
```
carbonin, line 457: Can we be consistent and either use `r.Credentials` or omit the `Credentials`?

I'd prefer it not be embedded FWIW (i.e. have a distinct field called `Credentials` like you did for `CertManager`)

internal/credentials/credentials.go
```diff
@@ -0,0 +1,257 @@
...
+		return nil, err
+	}
+	if existsAndValid {
+		return &clusterInfo.KubeconfigCryptoRetention, nil
+	}
+	// Generate cluster certs and create a new kubeconfig.
+	// TODO: handle user provided API and ingress certs
+	r.Log.Infof("Generating cluster crypto")
+	cm := certs.KubeConfigCertManager{}
+	if err := cm.GenerateAllCertificates(); err != nil {
```
carbonin, line 53: Do we ever want to call `GenerateAllCertificates` on an existing instance?
I feel like it would be safer for it to be `RegenerateAllCertificates` and throw out any existing ones at the start of the method.

carbonin (changes requested):

internal/credentials/credentials.go
eranco74: Sure, I don't think we need to change the name though.
GenerateAllCerts can just dump the existing ones

eranco74: Though when I think of it the existing certs is the only thing this object containes.
It dumping all the existing certs is essentially the same as creating a new object:
```
 func (r *KubeConfigCertManager) dumpExistingCertificates() {
	r = KubeConfigCertManager{}
}
```

carbonin: Sure, but it's just easier for a caller to create bugs if calling `GenerateAllCertificates` will append new certs.

I just don't see why that would ever be the desired behavior.

carbonin (approved):

PR #40, opened 2024-02-09
MGMT-16787: Allow ISO to be downloaded directly

carbonin:
Currently the only way to start the reconfiguration process of an IBI SNO is to attach the configuration ISO using a BareMetalHost.

There are cases where a user either does not want to expose the BMC across the network from the remote site to the hub cluster or is not using bare metal at all. For these cases the user needs an easy way to download the configuration image from this service.

This PR creates a route for the existing image download service and queries the route host on startup to be used as the base URL when images are created. When a new image is created the download URL is saved to the ImageClusterInstall status in the new `configurationImageURL` field.

Resolves https://issues.redhat.com/browse/MGMT-16787

Example resource with the new field:
```yaml
apiVersion: extensions.hive.openshift.io/v1alpha1
kind: ImageClusterInstall
metadata:
  name: ibi-test
  namespace: ibi-test
spec:
  clusterDeploymentRef:
    name: ibi-test
  clusterMetadata:
    adminKubeconfigSecretRef:
      name: ibi-test-admin-kubeconfig
    clusterID: 1cb699d0-716d-4511-891e-b13996443b30
    infraID: ibiotest-s9wm4
  hostname: ibitesthost
  imageSetRef:
    name: 4.14.10
  nodeIP: 192.168.111.25
status:
  conditions:
  - lastProbeTime: "2024-02-09T20:51:58Z"
    lastTransitionTime: "2024-02-09T20:49:19Z"
    message: Image is ready for use 
    reason: Ready
    status: "True"
    type: RequirementsMet
  - lastProbeTime: "2024-02-09T20:49:16Z"
    lastTransitionTime: "2024-02-09T20:49:16Z"
    status: Unknown
    type: Completed
  - lastProbeTime: "2024-02-09T20:49:16Z"
    lastTransitionTime: "2024-02-09T20:49:16Z"
    status: Unknown
    type: Failed
  - lastProbeTime: "2024-02-09T20:49:16Z"
    lastTransitionTime: "2024-02-09T20:49:16Z"
    status: "False"
    type: Stopped
  configurationImageURL: https://images-image-based-install-operator.apps.ostest.test.metalkube.org:443/images/ibi-test/ibi-test.iso

```


eranco74 (commented):
Does this mean that ironic will also get the image via the route instead of the service?

carbonin:
> Does this mean that ironic will also get the image via the route instead of the service?

Yes, but it doesn't seem to cause an issue and seemed simpler. This is also what we do in assisted FWIW.

PR #42, opened 2024-02-11
MGMT-16667: Validate pull secret

eranco74:
Validate pull secret
Pass the pullSecret as part of the cluster config instead of writing the secret to manifest dir

### DIFF
--- a/controllers/imageclusterinstall_controller.go
+++ b/controllers/imageclusterinstall_controller.go
@@ -17,7 +17,9 @@ limitations under the License.
 package controllers
 
 import (
+	"bytes"
 	"context"
+	"encoding/base64"
 	"encoding/json"
 	"fmt"
 	"net/url"
@@ -38,6 +40,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	k8serrors "k8s.io/apimachinery/pkg/util/errors"
 	utilrand "k8s.io/apimachinery/pkg/util/rand"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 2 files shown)


controllers/imageclusterinstall_controller.go
```diff
@@ -781,6 +761,62 @@ func generateInfraID(base string) string {
 	return fmt.Sprintf("%s-%s", base, utilrand.String(5))
 }
 
+// GetValidPullSecret validates the pull secret reference and format, return the pull secret data
+func (r *ImageClusterInstallReconciler) GetValidPullSecret(ctx context.Context, psRef *corev1.LocalObjectReference, namespace string) (string, error) {
```
carbonin, line 765: No reason for this to be exported (method should be lower case).

controllers/imageclusterinstall_controller.go
```diff
@@ -781,6 +761,62 @@ func generateInfraID(base string) string {
...
+
+	err := r.ValidatePullSecret(string(psData))
+	if err != nil {
+		return "", fmt.Errorf("invalid pull secret data in secret %w", err)
+	}
+	return string(psData), nil
+}
+
+// ImagePullSecret checks if the given string is a valid image pull secret and returns an error if not.
+func (r *ImageClusterInstallReconciler) ValidatePullSecret(pullSecret string) error {
```
carbonin, line 787: Same here, should be `validatePullSecret`

controllers/imageclusterinstall_controller.go
```diff
@@ -781,6 +761,62 @@ func generateInfraID(base string) string {
...
+	errs := []error{}
+
+	for d, a := range s.Auths {
+		_, authPresent := a["auth"]
+		_, credsStorePresent := a["credsStore"]
+		if !authPresent && !credsStorePresent {
+			errs = append(errs, fmt.Errorf("invalid pull secret: %q JSON-object requires either 'auth' or 'credsStore' field", d))
+			continue
+		}
+		data, err := base64.StdEncoding.DecodeString(a["auth"].(string))
```
carbonin, line 807: Is it valid for this not to be present? Might we end up with a panic here for `a["auth"]` being nil?

eranco74: It's validated few lines above, I'll update the code to use the same var

carbonin (approved):

PR #47, opened 2024-03-03
MGMT-16610: Add Proxy config option

tsorya:
Adding proxy param and noProxy required values.
Currently lca still doesn't support proxy as api param, this code creates basis for future usage

### DIFF
--- a/controllers/imageclusterinstall_controller_test.go
+++ b/controllers/imageclusterinstall_controller_test.go
@@ -1687,7 +1687,6 @@ var _ = Describe("proxy", func() {
 	)
 
 	BeforeEach(func() {
-
 		r = &ImageClusterInstallReconciler{
 			Client: nil,
 			Scheme: scheme.Scheme,


controllers/imageclusterinstall_controller.go
```diff
@@ -562,6 +562,9 @@ func (r *ImageClusterInstallReconciler) writeClusterInfo(ctx context.Context, lo
 		log.Infof("created new infra ID %s", infraID)
 	}
 
+	// TODO: add proxy to SeedReconfiguration when it will be supported
+	ici.Spec.Proxy = r.updateNoProxy(ici.Spec.Proxy, cd.Spec.ClusterName, cd.Spec.ClusterName)
```
eranco74, line 566: Instead of updating the CR spec,let's make this update on the proxy settings we write to the seedReconfig (once it's available).
Changing the CR  spec is discouraged

eranco74 (commented):
Overall it's great but let's wait until proxy is part of LCA API

controllers/imageclusterinstall_controller.go
tsorya: for sure, it was this way only cause we didn't have seed reconfiguration api yet

eranco74:
@tsoryayou now that https://github.com/openshift-kni/lifecycle-agent/pull/410 is merged you can update this PR

PR #51, opened 2024-03-22
MGMT-16931: Add an integration test

carbonin:
Running `make run-integration-test` will now deploy the operator, create the required CRDs, create the CRs needed for a simple test of the operator, and ensure the ISO can be downloaded.

https://issues.redhat.com/browse/MGMT-16931

### DIFF
--- a/hack/crds/clusterimagesets.yaml
+++ b/hack/crds/clusterimagesets.yaml
@@ -1,0 +1,60 @@
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: (devel)
+  labels:
+    operators.coreos.com/hive-operator.openshift-operators: ""
+  name: clusterimagesets.hive.openshift.io
+spec:
+  conversion:
+    strategy: None
+  group: hive.openshift.io
+  names:
+    kind: ClusterImageSet
+    listKind: ClusterImageSetList
+    plural: clusterimagesets
+    shortNames:
+    - imgset
+    singular: clusterimageset
+  scope: Cluster
+  versions:
+  - additionalPrinterColumns:
+    - jsonPath: .spec.releaseImage
+      name: Release
+      type: string
+    name: v1
+    schema:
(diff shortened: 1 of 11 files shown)


hack/crds/clusterimagesets.yaml
```diff
@@ -0,0 +1,60 @@
...
+      openAPIV3Schema:
+        description: ClusterImageSet is the Schema for the clusterimagesets API
+        properties:
+          apiVersion:
+            description: 'APIVersion defines the versioned schema of this representation
+              of an object. Servers should convert recognized schemas to the latest
+              internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
+            type: string
+          kind:
+            description: 'Kind is a string value representing the REST resource this
```
achuzhoy, line 37: nit: representing...represents in the same sentence. Maybe worth rephrasing. 
Also, while we try to break the lines 32 and 33, line 34 is a very long one.  Just raising it to your awareness. If you still wanna go ahead with it - ok with me.

hack/integration-test/wait-for-ready.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+condition="Ready"
+timeout="10m"
+namespace="image-based-install-operator"
+selector="app=image-based-install-operator"
+
+echo "Waiting for (pod) on namespace (${namespace}) with labels (${selector}) to be created..."
```
achuzhoy, line 8: nit: isn't it "in namespace", i.e. not on :)

hack/crds/clusterimagesets.yaml
carbonin: This is not actually something we own. This is taken directly from hive's CRD. Plus the `kind` description is very likely defined in kubernetes core code.

hack/integration-test/wait-for-ready.sh
carbonin: Yeah I can change this wording. I actually copied a bunch of this from assisted so I probably didn't review it closely enough :sweat_smile:

PR #52, opened 2024-04-05
MGMT-17379: Monitor the cluster installation after configuring a BMH

carbonin:
This checks the remote cluster for an available ClusterVersion and for
all the nodes to be Ready. If these conditions are not met by the
timeout (default 1h) the cluster will be marked as failed.

The timeout is configurable using an annotation on the
ImageClusterInstall (`imageclusterinstall.extensions.hive.openshift.io/install-timeout`)

Resolves https://issues.redhat.com/browse/MGMT-17379


carbonin:
Leaving this as WIP until I can test it live.

controllers/imageclusterinstall_controller.go
```diff
@@ -216,15 +225,126 @@ func (r *ImageClusterInstallReconciler) Reconcile(ctx context.Context, req ctrl.
 			return ctrl.Result{}, err
 		}
 
-		if err := r.setClusterInstalledConditions(ctx, ici); err != nil {
-			log.WithError(err).Error("failed to set installed conditions")
+		timedout, err := r.checkClusterTimeout(ctx, log, ici, r.DefaultInstallTimeout)
```
eranco74, line 228: We want to set the timeout only in case `clusterDeployment.Spec.Installed != true`.
If the controller already set it to true we don't want to keep querying the spoke (unsure what might trigger reconciliation after the `clusterDeployment.Spec.Installed` is set to `true`).

controllers/imageclusterinstall_controller.go
```diff
@@ -216,15 +225,126 @@ func (r *ImageClusterInstallReconciler) Reconcile(ctx context.Context, req ctrl.
...
+		if err != nil {
+			log.WithError(err).Error("failed to check for install timeout")
+			return ctrl.Result{}, err
+		}
+		if timedout {
+			return ctrl.Result{}, nil
```
eranco74, line 234: Add a log here

controllers/imageclusterinstall_controller.go
```diff
@@ -216,15 +225,126 @@ func (r *ImageClusterInstallReconciler) Reconcile(ctx context.Context, req ctrl.
...
+	}
+	installed, err := r.MonitorInstall(ctx, log, spokeClient)
+	if err != nil {
+		log.WithError(err).Error("failed to monitor spoke cluster")
+		return ctrl.Result{}, err
+	}
+	if !installed {
+		return ctrl.Result{RequeueAfter: time.Minute}, nil
+	}
+
```
eranco74, line 287: Add log here

controllers/imageclusterinstall_controller.go
```diff
@@ -216,15 +225,126 @@ func (r *ImageClusterInstallReconciler) Reconcile(ctx context.Context, req ctrl.
...
+	return false, nil
+}
+
+func (r *ImageClusterInstallReconciler) checkClusterStatus(ctx context.Context, log logrus.FieldLogger, ici *v1alpha1.ImageClusterInstall, clusterDeployment *hivev1.ClusterDeployment) (ctrl.Result, error) {
+	spokeClient, err := r.spokeClient(ctx, ici)
+	if err != nil {
+		log.WithError(err).Error("failed to create spoke client")
+		return ctrl.Result{}, err
+	}
+	installed, err := r.MonitorInstall(ctx, log, spokeClient)
```
eranco74, line 279: Cnosider renaming to:
 ```suggestion
	installed, err := r.GetSpokeClusterInstallStatus(ctx, log, spokeClient)
```
You can probably move the spoke client initialization into that function as well (it's the only one using it)

File: controllers/imageclusterinstall_controller.go [hunk 5]
```diff
@@ -216,15 +225,126 @@ func (r *ImageClusterInstallReconciler) Reconcile(ctx context.Context, req ctrl.
...
+		}
+
+		res, err := r.checkClusterStatus(ctx, log, ici, clusterDeployment)
+		if err != nil {
+			log.WithError(err).Error("failed to check cluster status")
 			return ctrl.Result{}, err
 		}
+		if !res.IsZero() {
+			return res, nil
+		}
```
eranco74, line 244: You can just `return res, nil` here

controllers/imageclusterinstall_controller.go
```diff
@@ -216,15 +225,126 @@ func (r *ImageClusterInstallReconciler) Reconcile(ctx context.Context, req ctrl.
...
+		log.WithError(err).Error("failed to create spoke client")
+		return ctrl.Result{}, err
+	}
+	installed, err := r.MonitorInstall(ctx, log, spokeClient)
+	if err != nil {
+		log.WithError(err).Error("failed to monitor spoke cluster")
+		return ctrl.Result{}, err
+	}
+	if !installed {
+		return ctrl.Result{RequeueAfter: time.Minute}, nil
```
eranco74, line 285: Consider adding log here

internal/monitor/monitor.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	isAvailable := false
+	for _, cond := range cv.Status.Conditions {
+		if cond.Type == configv1.OperatorAvailable {
+			if cond.Status == configv1.ConditionTrue {
+				isAvailable = true
+				break
+			} else {
+				log.Infof("cluster version is not yet available because %s: %s", cond.Reason, cond.Message)
```
eranco74, line 47: It would be nice to return this string and set it as the message on the installed condition.

internal/monitor/monitor.go
```diff
@@ -0,0 +1,78 @@
...
+		log.Info("no nodes found")
+		return false, nil
+	}
+
+	nodesReady := true
+	for _, node := range nodes.Items {
+		for _, cond := range node.Status.Conditions {
+			if cond.Type == corev1.NodeReady {
+				if cond.Status != corev1.ConditionTrue {
+					log.Infof("node %s is not yet ready", node.Name)
```
eranco74, line 70: It would be nice to return this string and set it as the message on the installed condition.

controllers/imageclusterinstall_controller.go
eranco74, line 285: Consider calling `setClusterInstalledConditions` here (set the condition to false with a message that reflects the spoke status)

controllers/imageclusterinstall_controller.go
carbonin: I'm fine with the rename, but I'd rather not move the spoke client stuff because this makes it a lot easier to mock the spoke client in the monitor code. Otherwise I need to create some kind of factory so I can mock it to return the fake client.

carbonin:
Should we be checking more?

When I tested this manually the cluster went to installed before all the cluster operators were ready.
Maybe we should also be looking at the `ClusterVersion` `Failing` condition?

```
    - lastTransitionTime: "2024-04-17T18:45:24Z"
      message: Cluster operator authentication is not available
      reason: ClusterOperatorNotAvailable
      status: "True"
      type: Failing
```

What do you think @tsorya ?

internal/monitor/monitor.go
```diff
@@ -0,0 +1,108 @@
...
+	clusterVersionAvailableMessage = "ClusterVersion is available"
+)
+
+type ClusterInstallStatus struct {
+	Installed            bool
+	ClusterVersionStatus string
+	NodesStatus          string
+}
+
+func (status *ClusterInstallStatus) String() string {
```
eranco74, line 27: :+1:

eranco74 (commented):
/lgtm

controllers/imageclusterinstall_controller.go
```diff
@@ -216,15 +225,131 @@ func (r *ImageClusterInstallReconciler) Reconcile(ctx context.Context, req ctrl.
...
+	}
+
+	if status := r.GetSpokeClusterInstallStatus(ctx, log, spokeClient); !status.Installed {
+		log.Infof("cluster install in progress: %s", status)
+		if err := r.setClusterInstallingConditions(ctx, ici, status.String()); err != nil {
+			log.WithError(err).Error("failed to set installing conditions")
+		}
+		return ctrl.Result{RequeueAfter: time.Minute}, nil
+	}
+	log.Info("cluster is installed")
```
tsorya, line 291: should we have some cluster id in logs? I mean anything that will allow to track what happened with specific cluster

internal/monitor/monitor.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	return ClusterInstallStatus{
+		Installed:            cvAvailable && nodesReady,
+		ClusterVersionStatus: cvMessage,
+		NodesStatus:          nodesMessage,
+	}
+}
+
+func clusterVersionStatus(ctx context.Context, log logrus.FieldLogger, c client.Client) (bool, string, error) {
```
tsorya, line 55: I wonder if this is a right way to go.
There is some point where cluster version didn't update its status yet i think but node will be ready already.
Maybe we should use some LCA CR to actually manage IBI installation? 
What i mean is that cluster version can even not to start yet when node is already ready, actually i think it will not start before node is ready. 
I think it has a pretty big chance to set installed before cluster actually starts all the pods.

controllers/imageclusterinstall_controller.go
carbonin: [This](https://github.com/openshift/image-based-install-operator/blob/52a99f434b241ef89d996bdff478e81fcd141d34/controllers/imageclusterinstall_controller.go#L113) is the first thing I do in the reconcile loop.

So the ImageClusterInstall name and namespace will always be printed as fields in every log message.
Looks like this:

`time="2024-04-17T18:45:23Z" level=info msg="cluster is installed" func="github.com/openshift/image-based-install-operator/controllers.(*ImageClusterInstallReconciler).checkClusterStatus" file="/workspace/controllers/imageclusterinstall_controller.go:299" name=ibi-test namespace=ibi-test`

internal/monitor/monitor.go
carbonin: Yeah, I wanted to ask you about this because I'm still seeing a lot of cluster operators not ready when these conditions are fulfilled.

Is there something better I can check or just go with this for now (it's better than it was for sure)

carbonin:
```
time="2024-04-19T15:38:13Z" level=info msg="Serving image for ImageClusterInstall ibi-test/ibi-test" func="github.com/openshift/image-based-install-operator/internal/imageserver.(*Handler).ServeHTTP" file="/workspace/internal/imageserver/imageserver.go:47"
time="2024-04-19T15:38:13Z" level=error msg="failed to acquire file lock" func="github.com/openshift/image-based-install-operator/internal/imageserver.(*Handler).ServeHTTP" file="/workspace/internal/imageserver/imageserver.go:73"
```

Interesting. Maybe we need to wait a bit after the image is set

PR #55, opened 2024-04-23
MGMT-17619: Align cluster install usage with hive expectations

carbonin:
~~Built on https://github.com/openshift/image-based-install-operator/pull/52~~ Merged
~~Requires https://github.com/openshift/hive/pull/2267~~ Merged

Hive will copy the cluster metadata, set the installed field, and will also only set conditions if the message or reason was updated.

Resolves https://issues.redhat.com/browse/MGMT-17619


carbonin:
Without https://github.com/openshift/hive/pull/2267 hive will still update the cluster deployment when provisioning finishes (successfully or not), but not while the process is ongoing.

PR #60, opened 2024-05-23
MGMT-17087: validate that bmh has at least one ip in provided machine

tsorya:
[MGMT-17087](https://issues.redhat.com//browse/MGMT-17087): validate that bmh has at least one ip in provided machine
[MGMT-17930](https://issues.redhat.com//browse/MGMT-17930): Skip bmh validation in case inspection annotation was set to disable
    If user disables bmh inspection we should skip bmh validations but in case it was not set we should wait till we get hw details and validate them


controllers/imageclusterinstall_controller.go
```diff
@@ -246,6 +266,49 @@ func (r *ImageClusterInstallReconciler) Reconcile(ctx context.Context, req ctrl.
...
+	}
+
+	clusterInfoFilePath, err := r.clusterInfoFilePath(ici)
+	if err != nil {
+		return err
+	}
+	clusterInfo := r.getClusterInfoFromFile(clusterInfoFilePath)
+
+	if bmh.Status.HardwareDetails == nil {
+		return fmt.Errorf("hardware details not found for BareMetalHost %s/%s", bmh.Namespace, bmh.Name)
```
carbonin, line 286: I know I said we'll rerun reconcile on an error, but this isn't really an error, right?
We know we're waiting on some external condition so I think it would be better to log an info and requeue here.

controllers/imageclusterinstall_controller_test.go
```diff
@@ -32,6 +32,21 @@ import (
 	. "github.com/onsi/gomega"
 )
 
+func getBHM(state bmh_v1alpha1.ProvisioningState) *bmh_v1alpha1.BareMetalHost {
```
carbonin, line 35: This should probably be `getBMH`, right?

carbonin: Maybe also just rename to `bmhInState` or something to avoid it sounding like this is querying for a BMH.

internal/utils/utils.go
```diff
@@ -0,0 +1,18 @@
+package utils
```
carbonin, line 1: Can we avoid a utils package?

If there's only one caller of this function just leave it with the caller.
If it needs to be shared we can break it out later.

controllers/imageclusterinstall_controller.go
tsorya: Actually i am not sure about it. Maybe lets log it but still set error? like Waiting for host hardware details?

eranco74: This should update the status to pending and add this error as the message + requeue

eranco74: We need a new condition, to what we have in ACI - https://github.com/openshift/assisted-service/blob/8a63485495f366dbdf45b33c31fd0994e0c0986a/api/hiveextension/v1beta1/agentclusterinstall_types.go#L33

api/v1alpha1/imageclusterinstall_types.go
```diff
@@ -41,6 +41,12 @@ const (
 
 	InstallSucceededReason  = "ClusterInstallationSucceeded"
 	InstallSucceededMessage = "Cluster installation has succeeded"
+
+	HostValidatedCondition     = "Validated"
+	HostValidationFailedReason = "HostValidationFailed"
+	HostValidationSucceeded    = "HostValidationSucceeded"
```
eranco74, line 47: Don't you need pending?

controllers/conditions.go
```diff
@@ -123,6 +129,25 @@ func (r *ImageClusterInstallReconciler) setHostConfiguredCondition(ctx context.C
 	return r.Status().Patch(ctx, ici, patch)
 }
 
+func (r *ImageClusterInstallReconciler) setHostValidationCondition(ctx context.Context, ici *v1alpha1.ImageClusterInstall, err error) error {
+	cond := hivev1.ClusterInstallCondition{
+		Type:    hivev1.ClusterInstallConditionType(v1alpha1.HostValidatedCondition),
+		Status:  corev1.ConditionTrue,
+		Reason:  v1alpha1.HostValidationSucceeded,
+		Message: v1alpha1.HostValidationsOKMsg,
+	}
```
eranco74, line 138: What does installation condition have to do with it?

controllers/imageclusterinstall_controller.go
```diff
@@ -246,6 +262,62 @@ func (r *ImageClusterInstallReconciler) Reconcile(ctx context.Context, req ctrl.
...
+	bmh *bmh_v1alpha1.BareMetalHost) (ctrl.Result, error) {
+
+	// no need to validate if inspect annotation is disabled
+	if bmh.ObjectMeta.Annotations != nil && bmh.ObjectMeta.Annotations[inspectAnnotation] == "disabled" {
+		r.Log.Infof("inspection is disabled for BareMetalHost %s/%s, skip hardware validation", bmh.Namespace, bmh.Name)
+		return ctrl.Result{}, nil
+	}
+
+	if bmh.Status.HardwareDetails == nil {
+		r.Log.Infof("hardware details not found for BareMetalHost %s/%s", bmh.Namespace, bmh.Name)
```
eranco74, line 277: This should set the validation condition to `pending`

controllers/imageclusterinstall_controller.go
```diff
@@ -246,6 +262,62 @@ func (r *ImageClusterInstallReconciler) Reconcile(ctx context.Context, req ctrl.
...
 }
 
+func (r *ImageClusterInstallReconciler) validateSeedReconfigurationWithBMH(
+	ctx context.Context,
+	ici *v1alpha1.ImageClusterInstall,
+	bmh *bmh_v1alpha1.BareMetalHost) (ctrl.Result, error) {
+
+	// no need to validate if inspect annotation is disabled
+	if bmh.ObjectMeta.Annotations != nil && bmh.ObjectMeta.Annotations[inspectAnnotation] == "disabled" {
+		r.Log.Infof("inspection is disabled for BareMetalHost %s/%s, skip hardware validation", bmh.Namespace, bmh.Name)
```
eranco74, line 272: This should set the validation condition to `success`/`skipped` with this message

api/v1alpha1/imageclusterinstall_types.go
tsorya: i assume we can leave it as unknown? Do we want actually to set pending?

controllers/conditions.go
tsorya: Same as with image ready one, this is type we use

controllers/imageclusterinstall_controller.go
tsorya: There is no such type as pending, we can set false though

controllers/conditions.go
```diff
@@ -71,6 +71,12 @@ func (r *ImageClusterInstallReconciler) initializeConditions(ctx context.Context
 			Reason:  "Unknown",
 			Message: "Unknown",
 		},
+		{
+			Type:    hivev1.ClusterInstallConditionType(v1alpha1.HostValidatedCondition),
```
carbonin, line 75: What's the difference between `Validated` and `RequirementsMet`?

If we can make this logic fit in one of the hive API conditions that would be better since it would then be copied to the cluster deployment.

tsorya: We use RequirementsMet to set Done for image, in this case it will be confusing to set RequirementsMet success and then after adding bmh reference to set it to failed. That's why i added new condition

carbonin: Do we really need a separate condition for each? I think it makes sense to set requirements met to success when we don't have a BMH, but this validation is also a good match for pre-install cluster requirements.

Or, put another way, we already would use a "validated" condition to mean a bunch of possible issues why not apply that same logic to the condition we already have?

carbonin: Also site config is only going to copy the hive contract conditions back to the site config status.

So those users will have to look at the ImageClusterInstall specifically to see this condition.

tsorya: To say the truth i have no objections, though it means that there will be some difference for same condition with and without bmh, but sounds like it is not a big deal

eranco74: The condition message can reflect the difference.

carbonin (approved):

PR #64, opened 2024-07-03
MGMT-18219 Use dataImage instead of BMH spec.image.URL

eranco74:
Update IBIO to use the new data image functionality introduced in OCP4.16 to deliver the configuration ISO and not to relay on the BMH image boot fallback.


eranco74:
Closing in favor of https://github.com/openshift/image-based-install-operator/pull/86

PR #70, opened 2024-08-14
ACM-13444: Update images to use registry.access.redhat.com bases

carbonin:
This also involved changing some of where the resulting executables were placed which then required a bundle update.

WIP until https://github.com/openshift/image-based-install-operator/pull/69 is merged.


carbonin:
It looks like this is now failing because of how the openshift CI is replacing base images during build.

Building this dockerfile as is deploys and runs fine for me locally.

Going to ask for some help from folks who understand openshift CI better than I do tomorrow.

carbonin:
Moved back to RHEL8 so we only solve one issue at a time.
Moving to RHEL9 is what caused the test failure because openshift/release still has the el8 image as the base image override.

PR #72, opened 2024-08-21
MGMT-18553: use same nmstate configuration logic as bmh

tsorya:
[MGMT-18553](https://issues.redhat.com//browse/MGMT-18553): In order to use same nmstate configuration logic as bmh we will use secret provided as reference in
PreprovisioningNetworkDataName in bmh spec, in case there is no bmh we will still use same configmap as before

### DIFF
--- a/controllers/imageclusterinstall_controller.go
+++ b/controllers/imageclusterinstall_controller.go
@@ -98,6 +98,7 @@ const (
 	extraManifestsDir            = "extra-manifests"
 	manifestsDir                 = "manifests"
 	nmstateCMKey                 = "network-config"
+	nmstateSecretKey             = "nmstate"
 	clusterInstallFinalizerName  = "imageclusterinstall." + v1alpha1.Group + "/deprovision"
 	caBundleFileName             = "tls-ca-bundle.pem"
 	imageBasedInstallInvoker     = "image-based-install"
@@ -167,7 +168,20 @@ func (r *ImageClusterInstallReconciler) Reconcile(ctx context.Context, req ctrl.
 		return ctrl.Result{}, err
 	}
 
-	if res, err := r.writeInputData(ctx, log, ici, clusterDeployment); !res.IsZero() || err != nil {
+	var bmh *bmh_v1alpha1.BareMetalHost
+	var err error
+	if ici.Spec.BareMetalHostRef != nil {
+		bmh, err = r.getBMH(ctx, ici.Spec.BareMetalHostRef)
+		if err != nil {
+			log.WithError(err).Error("failed to get BareMetalHost")
+			if updateErr := r.setHostConfiguredCondition(ctx, ici, err); updateErr != nil {
+				log.WithError(updateErr).Error("failed to update ImageClusterInstall status")
+			}
+			return ctrl.Result{}, err
+		}
+	}
+
+	if res, err := r.writeInputData(ctx, log, ici, clusterDeployment, bmh); !res.IsZero() || err != nil {
 		if err != nil {
 			if updateErr := r.setImageReadyCondition(ctx, ici, err, ""); updateErr != nil {
 				log.WithError(updateErr).Error("failed to update ImageClusterInstall status")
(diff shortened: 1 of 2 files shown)


controllers/imageclusterinstall_controller.go
```diff
@@ -723,7 +728,41 @@ func (r *ImageClusterInstallReconciler) imageSetRegistry(ctx context.Context, ic
 	return strings.Split(namedRef.Name(), "/")[0], nil
 }
 
+func (r *ImageClusterInstallReconciler) nmstateConfigFromBMH(ctx context.Context, ici *v1alpha1.ImageClusterInstall) (string, error) {
+	bmh, err := r.getBMH(ctx, ici.Spec.BareMetalHostRef)
```
carbonin, line 732: Can we pass the BMH through to this function instead of querying it multiple times?

controllers/imageclusterinstall_controller.go
```diff
@@ -159,14 +160,27 @@ func (r *ImageClusterInstallReconciler) Reconcile(ctx context.Context, req ctrl.
 			log.WithError(updateErr).Error("failed to update ImageClusterInstall status")
 			return ctrl.Result{}, updateErr
 		}
-		return ctrl.Result{Requeue: true}, nil
+		return ctrl.Result{RequeueAfter: 10 * time.Second}, nil
```
carbonin, line 163: Why change this?

controllers/imageclusterinstall_controller_test.go
```diff
@@ -496,6 +522,54 @@ var _ = Describe("Reconcile", func() {
 		Expect(json.Unmarshal(content, infoOut)).To(Succeed())
 		Expect(infoOut.RawNMStateConfig).To(Equal(validNMStateConfig))
 	})
+	It("copies the nmstate config bmh preprovisioningNetworkDataName even if congimap exists", func() {
```
carbonin, line 525: Should we also have a test that we will ignore the config in the ICI if nothing is set in the BMH?

That's the behavior we're expecting, right?

controllers/imageclusterinstall_controller.go
tsorya: In order not to Requeue immediately as this will create tons on logs

controllers/imageclusterinstall_controller.go
```diff
@@ -723,7 +728,41 @@ func (r *ImageClusterInstallReconciler) imageSetRegistry(ctx context.Context, ic
...
+	var nmstateData map[string]any
+	if err := yaml.Unmarshal(nmstate, &nmstateData); err != nil {
+		return "", fmt.Errorf("failed to unmarshal nmstate data: %w", err)
+	}
+
+	return string(nmstate), nil
+}
+
+// in case bmh was configured with static networking we want to use it
+// currently we added support for both configmap and secret but at some point we are gonna remove configmap support in order to use only secret
```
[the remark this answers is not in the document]
tsorya: Sorry i meant secret that bmh points to

controllers/imageclusterinstall_controller.go
tsorya: Actually as client has cache get command should cost nothing but code will be much cleaner, WDYT?

controllers/imageclusterinstall_controller.go
carbonin: Right, but I'm just making sure I understand. It's not actually directly related to this change, right?

tsorya: Sorry , yep was on the way , I can drop the change

carbonin: Is it maybe related to https://issues.redhat.com/browse/MGMT-18536?

tsorya: I think it was added when i looked on this ticket, let me remove this line :)

eranco74:
Now that users can generate the config iso with the openshift installer I think we can claim that IBIO is used for automation and not for generating config gISOs for manual installation. 
If the BMH ref will be mandatory we can remove the network configuration from the ICI CR and just get it from the BMH (if it's there).
I think we would want to do this ASAP (prior to GA).
@carbonin, @tsorya thoughts?

eranco74:
Users could still fake it by referencing a BMH CR with disabled inspections.

eranco74:
> We can remove non-bmh flow and remove configmap but it can be next PR

https://issues.redhat.com/browse/MGMT-18662

PR #74, opened 2024-08-22
ACM-13449: Label ConfigMaps and Secrets for backup

carbonin:
In order for the ACM hub DR solution to include our required ConfigMaps and Secrets, they must be labeled with
`cluster.open-cluster-management.io/backup=true`.

Ensure the label is present on each reconcile.

The ImageClusterInstall is already covered because all resources in the `hive.openshift.io` group are automatically backed up.

Resolves https://issues.redhat.com/browse/ACM-13449

### DIFF
--- a/controllers/imageclusterinstall_controller.go
+++ b/controllers/imageclusterinstall_controller.go
@@ -104,6 +104,8 @@ const (
 	invokerCMFileName            = "invoker-cm.yaml"
 	imageDigestMirrorSetFileName = "image-digest-sources.json"
 	installTimeoutAnnotation     = "imageclusterinstall." + v1alpha1.Group + "/install-timeout"
+	backupLabel                  = "cluster.open-cluster-management.io/backup"
+	backupLabelValue             = "true"
 )
 
 //+kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch;create;update;patch;delete
@@ -176,6 +178,8 @@ func (r *ImageClusterInstallReconciler) Reconcile(ctx context.Context, req ctrl.
 		}
 		return res, err
 	}
+
+	r.labelReferencedObjectsForBackup(ctx, log, ici, clusterDeployment)
 
 	imageUrl, err := url.JoinPath(r.BaseURL, "images", req.Namespace, fmt.Sprintf("%s.iso", req.Name))
 	if err != nil {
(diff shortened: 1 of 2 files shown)


controllers/imageclusterinstall_controller.go
```diff
@@ -574,6 +578,89 @@ func (r *ImageClusterInstallReconciler) removeBMHImage(ctx context.Context, bmhR
...
+	obj.SetLabels(labels)
+	return true
+}
+
+func (r *ImageClusterInstallReconciler) labelConfigMapForBackup(ctx context.Context, key types.NamespacedName) error {
+	cm := &corev1.ConfigMap{}
+	if err := r.Get(ctx, key, cm); err != nil {
+		return err
+	}
+
```
eranco74, line 600: Consider checking if the label already exists and return

controllers/imageclusterinstall_controller.go
```diff
@@ -574,6 +578,89 @@ func (r *ImageClusterInstallReconciler) removeBMHImage(ctx context.Context, bmhR
...
+	}
+	return nil
+}
+
+func (r *ImageClusterInstallReconciler) labelSecretForBackup(ctx context.Context, key types.NamespacedName) error {
+	secret := &corev1.Secret{}
+	if err := r.Get(ctx, key, secret); err != nil {
+		return err
+	}
+
```
eranco74, line 613: Consider checking if the label already exists and return

controllers/imageclusterinstall_controller.go
```diff
@@ -574,6 +578,89 @@ func (r *ImageClusterInstallReconciler) removeBMHImage(ctx context.Context, bmhR
...
+			log.WithError(err).Errorf("failed to label ConfigMap %s for backup", manifestKey)
+		}
+	}
+
+	if cd.Spec.PullSecretRef != nil {
+		psKey := types.NamespacedName{Name: cd.Spec.PullSecretRef.Name, Namespace: cd.Namespace}
+		if err := r.labelSecretForBackup(ctx, psKey); err != nil {
+			log.WithError(err).Errorf("failed to label Secret %s for backup", psKey)
+		}
+	}
```
eranco74, line 661: What about ImageSetRef and ClusterDeploymentRef?

controllers/imageclusterinstall_controller.go
carbonin: This is being done based on the return value of `setBackupLabel`.

I did it this way to share the logic between secrets and config maps.

controllers/imageclusterinstall_controller.go
carbonin: Same response as above.

controllers/imageclusterinstall_controller.go
carbonin: ClusterImageSets and ClusterDeployments are included in the backup by default.

carbonin:
Just FYI see https://docs.redhat.com/en/documentation/red_hat_advanced_cluster_management_for_kubernetes/2.10/html/business_continuity/business-cont-overview#resources-that-are-backed-up for the list of objects being backed up by default

eranco74 (commented):
/lgtm
/approve

PR #75, opened 2024-08-26
MGMT-18536: We update annotation each time bmh was updated this causes reconcile storm

tsorya:
[MGMT-18536](https://issues.redhat.com//browse/MGMT-18536): We update annotation each time bmh was updated this causes reconcile storm
In this fix we will validate if annotation was set already


oourfali:
CC also @carbonin and @eranco74

eranco74 (commented):
/lgtm
/approve

PR #76, opened 2024-08-28
MGMT-18660: invalidating ironic image cache by removing image from bmh on data change

tsorya:
[MGMT-18660](https://issues.redhat.com//browse/MGMT-18660): invalidating ironic image cache by removing image from bmh on data change
In order to cleanup bmh image we should be sure that change occured
1. Write seed reconfiguration file only in case there is a change.
2. Stop rehashing kubeadmin password all the time but rather do it once
3. Verify hash of the configuration folder before and after going through all the params


controllers/imageclusterinstall_controller.go
```diff
@@ -207,21 +210,32 @@ func (r *ImageClusterInstallReconciler) Reconcile(ctx context.Context, req ctrl.
 
 	// in case there is no bmh we should set requirements met condition to true with image ready message
 	// in case bmh was set we will set this condition after host validations
-	if ici.Status.BareMetalHostRef == nil {
+	if ici.Spec.BareMetalHostRef == nil {
```
carbonin, line 213: I guess this was a typo originally? Not sure why else this would be changing here.

controllers/imageclusterinstall_controller.go
```diff
@@ -207,21 +210,32 @@ func (r *ImageClusterInstallReconciler) Reconcile(ctx context.Context, req ctrl.
...
 	}
 
 	if bmh != nil {
+		// in case image data was changed and bmh has image url configured
+		// we should remove image from it in order to invalidate ironic cache
+		if bmh.Spec.Image != nil && updated {
+			r.Log.Info("Image data was changed, removing image from BareMetalHost")
+			removed, err := r.removeBMHImage(ctx, ici.Spec.BareMetalHostRef)
+			if err != nil || removed {
+				// in case of error log will be printed by reconcile loop
```
carbonin, line 234: We are in `Reconcile` currently, who is printing this log?

internal/credentials/credentials_test.go
```diff
@@ -139,18 +139,20 @@ var _ = Describe("Credentials", func() {
...
-		passwordHash, err := cm.EnsureAdminPasswordSecret(ctx, clusterDeployment)
+		passwordHash, err := cm.EnsureAdminPasswordSecret(ctx, clusterDeployment, "")
 		Expect(err).NotTo(HaveOccurred())
 		// Verify the password secret
 		passwordSecret := &corev1.Secret{}
 		err = cm.Client.Get(ctx, client.ObjectKey{Namespace: clusterDeployment.Namespace, Name: clusterDeployment.Name + "-admin-password"}, passwordSecret)
 		Expect(err).NotTo(HaveOccurred())
 		password, exists := passwordSecret.Data["password"]
 		Expect(exists).To(BeTrue())
-		// Call again, the password hash should change but the password shouldn't
```
carbonin, line 150: Is there any reason we wanted this behavior previously?
@eranco74 ?

controllers/imageclusterinstall_controller.go
```diff
@@ -753,35 +773,57 @@ func (r *ImageClusterInstallReconciler) writeInputData(
...
 			return fmt.Errorf("failed to ensure admin password secret: %w", err)
 		}
-		if err := r.writeClusterInfo(ctx, log, ici, cd, crypto, psData, kubeadminPasswordHash, clusterInfoFilePath, clusterInfo, bmh); err != nil {
-			return fmt.Errorf("failed to write cluster info: %w", err)
+
+		newClusterInfo, err := r.createSeedReconfiguration(ctx, log, ici, cd, crypto, psData, kubeadminPasswordHash, clusterInfo, bmh)
+		if err != nil {
+			return fmt.Errorf("failed to create seed reconfiguration: %w", err)
 		}
+		if !reflect.DeepEqual(newClusterInfo, clusterInfo) {
```
carbonin, line 794: Do we really need this check if we're checking the directory hash later?

controllers/imageclusterinstall_controller_test.go
```diff
@@ -903,20 +895,101 @@ var _ = Describe("Reconcile", func() {
...
+		res, err = r.Reconcile(ctx, req)
+		Expect(err).NotTo(HaveOccurred())
+		res, err = r.Reconcile(ctx, req)
+		Expect(err).NotTo(HaveOccurred())
+		res, err = r.Reconcile(ctx, req)
+		Expect(err).NotTo(HaveOccurred())
+		res, err = r.Reconcile(ctx, req)
+		Expect(err).NotTo(HaveOccurred())
+		res, err = r.Reconcile(ctx, req)
+		Expect(err).NotTo(HaveOccurred())
```
carbonin, line 934: This is a bit excessive I think

carbonin (commented):
No need to do it now, but your PR description lists three individual changes you made. Those would make for awesome commits since we don't squash PR commits in this repo

controllers/imageclusterinstall_controller.go
tsorya: Yeah, i believe we should check Spec and not status here

controllers/imageclusterinstall_controller.go
tsorya: In case error occur manager prints the error

controllers/imageclusterinstall_controller.go
tsorya: I believe we can remove it though what's the point to write file if we don't need but we do it for many other cases, so ill remove it

controllers/imageclusterinstall_controller_test.go
tsorya: oops :) actually by mistake, it was part of my manual testing

controllers/imageclusterinstall_controller.go
```diff
@@ -753,35 +773,51 @@ func (r *ImageClusterInstallReconciler) writeInputData(
...
 			return fmt.Errorf("failed to ensure admin password secret: %w", err)
 		}
-		if err := r.writeClusterInfo(ctx, log, ici, cd, crypto, psData, kubeadminPasswordHash, clusterInfoFilePath, clusterInfo, bmh); err != nil {
+
+		newClusterInfo, err := r.createSeedReconfiguration(ctx, log, ici, cd, crypto, psData, kubeadminPasswordHash, clusterInfo, bmh)
+		if err != nil {
+			return fmt.Errorf("failed to create seed reconfiguration: %w", err)
+		}
+
+		if err := r.writeClusterInfo(newClusterInfo, clusterInfoFilePath); err != nil {
```
carbonin, line 795: We probably now don't need to split these, right?

And if you want to optimize out the file write if nothing changed then you can do the deep equal check in the function before writing it out, right?

tsorya: yep as it was before

internal/credentials/credentials_test.go
eranco74: What do you mean?
What behavior?

carbonin: Right, but we didn't *need* the hash to change for any reason, right?

controllers/imageclusterinstall_controller.go
carbonin: I mean you can recombine `createSeedReconfiguration` and `writeClusterInfo` into a single function. Is there any reason for them to be separate?

tsorya: ooo, right , when i removed the logic i forgot that i split it

internal/credentials/credentials_test.go
tsorya: We talked with @eranco74 today and he didn't remember any reason for it

carbonin (approved):

PR #85, opened 2024-09-05
MGMT-18834: validate before each patch if something changed before running patch itself.

tsorya:
[MGMT-18834](https://issues.redhat.com//browse/MGMT-18834): validate before each patch if something changed before running patch itself.
Add logging for each patch operation

### DIFF
--- a/controllers/conditions.go
+++ b/controllers/conditions.go
@@ -162,24 +162,28 @@ func (r *ImageClusterInstallReconciler) setRequirementsMetCondition(ctx context.
 
 func (r *ImageClusterInstallReconciler) setClusterInstalledConditions(ctx context.Context, ici *v1alpha1.ImageClusterInstall) error {
 	patch := client.MergeFrom(ici.DeepCopy())
-	setClusterInstallCondition(&ici.Status.Conditions, hivev1.ClusterInstallCondition{
+	completedUpdated := setClusterInstallCondition(&ici.Status.Conditions, hivev1.ClusterInstallCondition{
 		Type:    hivev1.ClusterInstallCompleted,
 		Status:  corev1.ConditionTrue,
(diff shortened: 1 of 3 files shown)


controllers/conditions.go
```diff
@@ -158,6 +181,7 @@ func (r *ImageClusterInstallReconciler) setClusterInstalledConditions(ctx contex
 		Message: v1alpha1.InstallSucceededMessage,
 	})
 
+	r.Log.Info("Setting cluster installed conditions")
```
carbonin, line 184: Shouldn't we check the return of these `setClusterInstallCondition` calls?

controllers/conditions.go
```diff
@@ -183,29 +207,34 @@ func (r *ImageClusterInstallReconciler) setClusterTimeoutConditions(ctx context.
 		Message: message,
 	})
 
+	r.Log.Info("Setting cluster timeout conditions")
```
carbonin, line 210: Same here, shouldn't this patch and log be conditional on the calls above?

controllers/conditions.go
tsorya: not in this case as we will set this condition once as reconcile will exit in case it was set already

controllers/conditions.go
tsorya: not in this case as we will set this condition once as reconcile will exit in case it was set already

carbonin (approved):

controllers/conditions.go
eranco74: And yet, let's check and not relay on some other code to returnearly, this is exactly what this pr should do.

tsorya: why? there is not issue with it? We should not overcomplicate each part of the the code, there is a unitest that check that we don't run after setting installed

PR #86, opened 2024-09-05
MGMT-18219: Use DataImage instead of BMH spec.image.URL

eliorerz:
/cc @eliorerz

### DIFF
--- a/controllers/imageclusterinstall_controller.go
+++ b/controllers/imageclusterinstall_controller.go
@@ -39,6 +39,7 @@ import (
 	"gopkg.in/yaml.v3"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
@@ -96,6 +97,7 @@ const (
 	detachedAnnotation           = "baremetalhost.metal3.io/detached"
 	detachedAnnotationValue      = "imageclusterinstall-controller"
 	inspectAnnotation            = "inspect.metal3.io"
+	rebootAnnotation             = "reboot.metal3.io"
 	clusterConfigDir             = "cluster-configuration"
 	extraManifestsDir            = "extra-manifests"
 	manifestsDir                 = "manifests"
@@ -118,9 +120,11 @@ const (
 //+kubebuilder:rbac:groups=extensions.hive.openshift.io,resources=imageclusterinstalls/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=extensions.hive.openshift.io,resources=imageclusterinstalls/finalizers,verbs=update
 //+kubebuilder:rbac:groups=metal3.io,resources=baremetalhosts,verbs=get;list;watch;update;patch
+//+kubebuilder:rbac:groups=metal3.io,resources=baremetalhosts/finalizers,verbs=update
 //+kubebuilder:rbac:groups=hive.openshift.io,resources=clusterdeployments,verbs=get;list;watch;update;patch
 //+kubebuilder:rbac:groups=hive.openshift.io,resources=clusterdeployments/finalizers,verbs=update
 //+kubebuilder:rbac:groups=hive.openshift.io,resources=clusterimagesets,verbs=get;list;watch
+//+kubebuilder:rbac:groups=metal3.io,resources=dataimages,verbs=get;list;watch;create;update;patch;delete
 
 func (r *ImageClusterInstallReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	log := r.Log.WithFields(logrus.Fields{"name": req.Name, "namespace": req.Namespace})
@@ -191,7 +195,7 @@ func (r *ImageClusterInstallReconciler) Reconcile(ctx context.Context, req ctrl.
 		}
 	}
 
-	res, updated, err := r.writeInputData(ctx, log, ici, clusterDeployment, bmh)
+	res, _, err := r.writeInputData(ctx, log, ici, clusterDeployment, bmh)
 	if !res.IsZero() || err != nil {
 		if err != nil {
 			if updateErr := r.setImageReadyCondition(ctx, ici, err, ""); updateErr != nil {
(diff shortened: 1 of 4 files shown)


tsorya:
lets wait till all the fixes for scale tests will be pushed

controllers/imageclusterinstall_controller_test.go
```diff
@@ -823,7 +823,7 @@ var _ = Describe("Reconcile", func() {
 	})
 
 	It("configures a referenced BMH", func() {
-		bmh := bmhInState(bmh_v1alpha1.StateAvailable)
+		bmh := bmhInState(bmh_v1alpha1.StateExternallyProvisioned)
```
carbonin, line 826: Are we expecting people to configure the BMH with externally provisioned now?
Will that disrupt inspection?

carbonin (commented):
Generally looks good.
Did you run through an e2e test with this? Specifically I'd want to make sure that inspection still works as expected.

controllers/imageclusterinstall_controller.go
```diff
@@ -521,27 +524,55 @@ func (r *ImageClusterInstallReconciler) SetupWithManager(mgr ctrl.Manager) error
...
-		bmh.Spec.Image = &bmh_v1alpha1.Image{}
-		dirty = true
-	}
-	if bmh.Spec.Image.URL != url {
-		bmh.Spec.Image.URL = url
-		dirty = true
-	}
-	liveIso := "live-iso"
-	if bmh.Spec.Image.DiskFormat == nil || *bmh.Spec.Image.DiskFormat != liveIso {
-		bmh.Spec.Image.DiskFormat = &liveIso
```
eranco74, line 542: How does this change affect the BMH status?
Will the BMH move to provisioned or stay available?

controllers/imageclusterinstall_controller.go
```diff
@@ -577,30 +608,22 @@ func (r *ImageClusterInstallReconciler) getBMH(ctx context.Context, bmhRef *v1al
...
-			return false, err
+	if err := r.Get(ctx, key, dataImage); err != nil {
+		if apierrors.IsNotFound(err) {
+			return true, nil
 		}
+		return false, err
 	}
 
-	return dirty, nil
+	return true, r.Client.Delete(ctx, dataImage)
```
eranco74, line 626: Same here, what happens to the BMH when we remove the data image?
* I'm guessing nothing changes

controllers/imageclusterinstall_controller_test.go
eliorerz: Yes, when the user configures the BMH with externally provisioned inspection will be disabled

controllers/imageclusterinstall_controller.go
```diff
@@ -310,6 +313,10 @@ func (r *ImageClusterInstallReconciler) validateSeedReconfigurationWithBMH(
 	ici *v1alpha1.ImageClusterInstall,
 	bmh *bmh_v1alpha1.BareMetalHost) (ctrl.Result, error) {
 
+	if bmh.Spec.ExternallyProvisioned {
+		return ctrl.Result{}, nil
```
eranco74, line 317: Add:
```suggestion
		msg := fmt.Sprintf("BareMetalHost %s/%s, is externally provisioned, skipping hardware validation", bmh.Namespace, bmh.Name)
		r.Log.Info(msg)
		if updateErr := r.setRequirementsMetCondition(ctx, ici, corev1.ConditionTrue, v1alpha1.HostValidationSucceeded, msg); updateErr != nil {
			r.Log.WithError(updateErr).Error("failed to update ImageClusterInstall status")
		}
		return ctrl.Result{}, nil
```

controllers/imageclusterinstall_controller.go
```diff
@@ -320,6 +327,16 @@ func (r *ImageClusterInstallReconciler) validateSeedReconfigurationWithBMH(
 		return ctrl.Result{}, nil
 	}
 
+	if bmh.Status.Provisioning.State != bmh_v1alpha1.StateAvailable {
+		r.Log.Infof("BareMetalHost (%s/%s) isn't available", bmh.Namespace, bmh.Name)
+		msg := fmt.Sprintf("BareMetalHost (%s/%s) provisioning state is: %s", bmh.Namespace, bmh.Name, bmh.Status.Provisioning.State)
```
eranco74, line 332:
```suggestion
		msg := fmt.Sprintf("BareMetalHost (%s/%s) provisioning state is: %s, waiting for %s", bmh.Namespace, bmh.Name, bmh.Status.Provisioning.State, bmh_v1alpha1.StateAvailable)
```

controllers/imageclusterinstall_controller.go
```diff
@@ -310,6 +313,10 @@ func (r *ImageClusterInstallReconciler) validateSeedReconfigurationWithBMH(
 	ici *v1alpha1.ImageClusterInstall,
 	bmh *bmh_v1alpha1.BareMetalHost) (ctrl.Result, error) {
 
+	if bmh.Spec.ExternallyProvisioned {
+		return ctrl.Result{}, nil
+	}
+
 	// no need to validate if inspect annotation is disabled
 	if bmh.ObjectMeta.Annotations != nil && bmh.ObjectMeta.Annotations[inspectAnnotation] == "disabled" {
```
eranco74, line 321: Use isInspectionEnabled instead

controllers/imageclusterinstall_controller.go
```diff
@@ -537,41 +554,101 @@ func (r *ImageClusterInstallReconciler) SetupWithManager(mgr ctrl.Manager) error
...
+	}
+	err := r.Get(ctx, key, &dataImage)
+	if apierrors.IsNotFound(err) {
+		return nil
+	}
+
+	return &dataImage
+}
+
+func (r *ImageClusterInstallReconciler) isInspectionEnabled(bmh *bmh_v1alpha1.BareMetalHost) bool {
```
eranco74, line 574:
```suggestion
func isInspectionEnabled(bmh *bmh_v1alpha1.BareMetalHost) bool {
```

controllers/imageclusterinstall_controller.go
```diff
@@ -537,41 +554,101 @@ func (r *ImageClusterInstallReconciler) SetupWithManager(mgr ctrl.Manager) error
...
+	}
+
+	if bmh.ObjectMeta.Annotations != nil && bmh.ObjectMeta.Annotations[inspectAnnotation] != "disabled" {
+		return true
+	}
+
+	return false
+}
+
+func (r *ImageClusterInstallReconciler) createBMHDataImage(ctx context.Context, bmh *bmh_v1alpha1.BareMetalHost, ici *v1alpha1.ImageClusterInstall, url string) error {
```
eranco74, line 586: Consider splitting into 2 functions:
1. createBMHDataImage
2. updateBMHProvisionigState

controllers/imageclusterinstall_controller.go
```diff
@@ -537,41 +554,101 @@ func (r *ImageClusterInstallReconciler) SetupWithManager(mgr ctrl.Manager) error
...
+			Spec: bmh_v1alpha1.DataImageSpec{
+				URL: url,
+			},
+		}
+		err := controllerutil.SetControllerReference(bmh, dataImage, r.Client.Scheme())
+		if err != nil {
+			return fmt.Errorf("failed to set controller reference for dataImage due to %w", err)
+		}
+
+		err = r.Create(ctx, dataImage)
```
eranco74, line 607: Should we keep a ref to the dataImage on the ICI?

controllers/imageclusterinstall_controller.go
```diff
@@ -310,6 +313,10 @@ func (r *ImageClusterInstallReconciler) validateSeedReconfigurationWithBMH(
 	ici *v1alpha1.ImageClusterInstall,
 	bmh *bmh_v1alpha1.BareMetalHost) (ctrl.Result, error) {
 
+	if bmh.Spec.ExternallyProvisioned {
```
tsorya, line 316: maybe lets add comment that there is nothing to validate in case host is ExternallyProvisioned as it will not be inspected?

controllers/imageclusterinstall_controller.go
```diff
@@ -228,20 +232,19 @@ func (r *ImageClusterInstallReconciler) Reconcile(ctx context.Context, req ctrl.
...
-			bmh.Status.Provisioning.State != bmh_v1alpha1.StateProvisioning &&
-			updated {
-			r.Log.Info("Image data was changed, removing image from BareMetalHost")
-			removed, err := r.removeBMHImage(ctx, ici.Spec.BareMetalHostRef)
+		// In case data image was changed and bmh is not externally provisioned yet,
+		// we should remove the DataImage from it in order to invalidate ironic cache
+		dataImage := r.getBmhDataImage(ctx, bmh)
+		if dataImage != nil && bmh.Status.Provisioning.State != bmh_v1alpha1.StateExternallyProvisioned && updated {
+			r.Log.Info("Data image was changed, removing data image")
+			removed, err := r.removeBMHDataImage(ctx, bmh, ici.Spec.BareMetalHostRef)
```
eranco74, line 247: IBIO set the BMH provisioning state to externallyProvisioned right after we create the dataImage.
Do we need this code? 
I see it was here, it's just not clear what the scenario + we know that it doesn't actually work (the data image won't get deleted)

controllers/imageclusterinstall_controller.go
```diff
@@ -537,41 +554,101 @@ func (r *ImageClusterInstallReconciler) SetupWithManager(mgr ctrl.Manager) error
...
 		bmh.Spec.AutomatedCleaningMode = bmh_v1alpha1.CleaningModeDisabled
+		r.Log.Infof("Disable automated cleaning mode for BareMetalHost (%s/%s)", bmh.Name, bmh.Namespace)
 		dirty = true
 	}
 
-	if bmh.Status.Provisioning.State == bmh_v1alpha1.StateProvisioned {
+	// boot from disk
+	if bmh.Status.Provisioning.State == bmh_v1alpha1.StateExternallyProvisioned && bmh.Status.PoweredOn {
 		if setAnnotaitonIfNotExists(&bmh.ObjectMeta, detachedAnnotation, detachedAnnotationValue) {
+			r.Log.Infof("Adding detached annotations to BareMetalHost (%s/%s)", bmh.Name, bmh.Namespace)
```
eranco74, line 645: Wait, won't this prevent ironic from performing the reboot?

controllers/imageclusterinstall_controller.go
eranco74, line 607: Log that we created the data image (add namespace + name in the log)

controllers/imageclusterinstall_controller.go
```diff
@@ -537,41 +554,101 @@ func (r *ImageClusterInstallReconciler) SetupWithManager(mgr ctrl.Manager) error
...
 	}
-	liveIso := "live-iso"
-	if bmh.Spec.Image.DiskFormat == nil || *bmh.Spec.Image.DiskFormat != liveIso {
-		bmh.Spec.Image.DiskFormat = &liveIso
+
+	if bmh.Status.Provisioning.State == bmh_v1alpha1.StateAvailable && bmh.Status.PoweredOn &&
+		setAnnotaitonIfNotExists(&bmh.ObjectMeta, rebootAnnotation, "true") {
+		r.Log.Infof("Adding reboot annotations to BareMetalHost (%s/%s)", bmh.Name, bmh.Namespace)
 		dirty = true
 	}
```
eranco74, line 634: I think what we need is this:
```suggestion
	if bmh.Status.PoweredOn{
		setAnnotaitonIfNotExists(&bmh.ObjectMeta, rebootAnnotation, "true") {
		r.Log.Infof("Adding reboot annotations to BareMetalHost (%s/%s)", bmh.Name, bmh.Namespace)
		dirty = true
	}
	if !bmh.Spec.online{
	        bmh.Spec.online = true
		r.Log.Infof("Setting BareMetalHost (%s/%s)  online to true", bmh.Name, bmh.Namespace)
		dirty = true
	}
```

controllers/imageclusterinstall_controller.go
tsorya: Scenario is like this :
1. You set wrong machine cidr
2. Validation failed.
3. You change ici
4. Invalidate cache

controllers/imageclusterinstall_controller.go
tsorya: Why? I don't think you are right here, you should not reboot node all the time bmh.Status.PoweredOn is set to true. You will never start installation. 
We should reboot node only in very strictly designed flows, like here.
We reboot only in case node had inspection, passed validation and is powered on, so we want to reboot into disk.
Maybe we are missing the case were node was provided powered on and we want to reboot it with new data image  but we should not look only on PoweredOn flag for sure.
I believe this is optimization and can be done after , for sure better not to reboot when we should not

controllers/imageclusterinstall_controller.go
tsorya: It should actually, when we started the installation we should not allow reboots

controllers/imageclusterinstall_controller.go
eranco74: So please add a comment that the code handles this case and make sure we set externally provisioned when when doing this.

controllers/imageclusterinstall_controller.go
eranco74: So let's move this to when the installation is complete

controllers/imageclusterinstall_controller.go
```diff
@@ -537,41 +554,101 @@ func (r *ImageClusterInstallReconciler) SetupWithManager(mgr ctrl.Manager) error
...
 
 	if bmh.Spec.AutomatedCleaningMode != bmh_v1alpha1.CleaningModeDisabled {
 		bmh.Spec.AutomatedCleaningMode = bmh_v1alpha1.CleaningModeDisabled
+		r.Log.Infof("Disable automated cleaning mode for BareMetalHost (%s/%s)", bmh.Name, bmh.Namespace)
```
eranco74, line 638: We need to move this to happen earlier

controllers/imageclusterinstall_controller.go
eliorerz: Why, the name and namespace are equal to the BMH, if we have bmh ref I don't think we need also a dataImage ref

controllers/imageclusterinstall_controller.go
eranco74: Seems that we should wrap the logic that reboots the BMH with a check of a condition that we set when we trigger the reboot (installation started...)

controllers/imageclusterinstall_controller.go
eliorerz: Why? Isn't taht the case we want to reboot the host?

controllers/imageclusterinstall_controller.go
eliorerz: We do, see few lines above
```go
r.Log.Infof("creating new dataImage for BareMetalHost (%s/%s)", bmh.Name, bmh.Namespace)
```

controllers/imageclusterinstall_controller.go
eliorerz: I'm not sure about why to remove the state available condition?
We are handling the online flag few lines above this part regardless the power on status, so here it already should be enabled.

controllers/imageclusterinstall_controller.go
eliorerz: WDYM by earlier?

tsorya: This is should be the first thing we do when we see bmh, i assume lets move it to another function and run it as first thing in the loop

controllers/imageclusterinstall_controller.go
tsorya: We actually want bmh not to be rebooted when installation finished , i believe we should set it right after we started installation but setting it after checkStatus should be ok too

controllers/imageclusterinstall_controller.go
eranco74: if validation fails you shouldn't create the data image so there's nothing to invalidate.

Is there a valid scenario for this?

controllers/imageclusterinstall_controller.go
tsorya: Opened a ticket for next pr https://issues.redhat.com/browse/MGMT-18899, for this specific case

controllers/imageclusterinstall_controller.go
tsorya: I believe we can remove, yeah missed that we set it after it. Though we need to create scenario that allows to update ici and clean the cache

controllers/imageclusterinstall_controller.go
```diff
@@ -96,6 +97,7 @@ const (
 	detachedAnnotation           = "baremetalhost.metal3.io/detached"
 	detachedAnnotationValue      = "imageclusterinstall-controller"
 	inspectAnnotation            = "inspect.metal3.io"
+	rebootAnnotation             = "metal3.io/reboot"
```
tsorya, line 100: reboot.metal3.io

PR #91, opened 2024-09-08
MGMT-18860: create ibio start cm

tsorya:
MGMT-18860: IBIO report installtion complete although the cluster is still reconfiguring
    Create new cm in order to get creation timestamp that will point us to
    the time when cluster started to run
    Check conditions in CVO and verify that at least one was changed after
    cm creation timestamp

### DIFF
--- a/controllers/imageclusterinstall_controller.go
+++ b/controllers/imageclusterinstall_controller.go
@@ -760,6 +760,9 @@ func (r *ImageClusterInstallReconciler) writeInputData(
 		if err := r.writeInvokerCM(filepath.Join(manifestsPath, invokerCMFileName)); err != nil {
 			return fmt.Errorf("failed to write invoker config map: %w", err)
 		}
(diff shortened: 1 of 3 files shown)


controllers/imageclusterinstall_controller.go
```diff
@@ -1263,6 +1266,27 @@ func (r *ImageClusterInstallReconciler) writeInvokerCM(filePath string) error {
...
 }
 
+func (r *ImageClusterInstallReconciler) writeIBIOStartTimeCM(filePath string) error {
+	cm := &corev1.ConfigMap{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: corev1.SchemeGroupVersion.String(),
+			Kind:       "ConfigMap",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: monitor.OcpConfigNamespace,
```
eranco74, line 1276: Consider using `openshift-lifecycle-agent` NS
openshift-config has nothing to do with IBI

eranco74 (commented):
Just a nit

eranco74 (commented):
/lgtm

PR #92, opened 2024-09-10
MGMT-18619: Obfuscate config ISO url

eranco74:
Use the ICI.UID instead of the ICI.name

### DIFF
--- a/controllers/imageclusterinstall_controller.go
+++ b/controllers/imageclusterinstall_controller.go
@@ -35,7 +35,6 @@ import (
 	_ "crypto/sha256"
 	_ "crypto/sha512"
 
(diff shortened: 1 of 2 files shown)


controllers/imageclusterinstall_controller.go
```diff
@@ -54,6 +53,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
+	"github.com/sirupsen/logrus"
+
```
tsorya, line 57: delete space?

eranco74: God knows what goimport was doing :man_shrugging:

PR #94, opened 2024-09-11
MGMT-18662 Remove static netowrk ref from the ICI

eranco74:
Following this change static network config is supported only via PreprovisioningNetworkDataName on the BMH spec

This should reduce the code complexity and confusion e.g. https://issues.redhat.com/browse/CNF-14123

### DIFF
--- a/controllers/imageclusterinstall_controller.go
+++ b/controllers/imageclusterinstall_controller.go
@@ -674,13 +674,6 @@ func (r *ImageClusterInstallReconciler) labelReferencedObjectsForBackup(ctx cont
 		}
 	}
 
-	if ici.Spec.NetworkConfigRef != nil {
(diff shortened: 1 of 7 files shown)


controllers/imageclusterinstall_controller.go
```diff
@@ -875,7 +868,11 @@ func (r *ImageClusterInstallReconciler) imageSetRegistry(ctx context.Context, ic
 	return strings.Split(namedRef.Name(), "/")[0], nil
 }
 
-func (r *ImageClusterInstallReconciler) nmstateConfigFromBMH(ctx context.Context, bmh *bmh_v1alpha1.BareMetalHost) (string, error) {
+// in case bmh was configured with static networking we want to use it's configuration
```
tsorya, line 871: Change comment?

eranco74: Why? the comment is currect

PR #95, opened 2024-09-12
MGMT-18876: Update admission controller to block changes to ICI once the installation started

eranco74:
This change also removes the handleing for BMH update post install since the effect of such update is unknown
Note that the webhook doesn't handle changes to resources referenced by the ICI

### DIFF
--- a/api/v1alpha1/imageclusterinstall_webhook.go
+++ b/api/v1alpha1/imageclusterinstall_webhook.go
@@ -53,21 +53,6 @@ func (r *ImageClusterInstall) ValidateCreate() (admission.Warnings, error) {
 	return nil, nil
 }
 
-func (r *ImageClusterInstall) validate() error {
-	if err := isValidSSHPublicKey(r.Spec.SSHKey); err != nil {
-		return fmt.Errorf("invalid ssh key: %v", err)
-	}
-	if err := isValidHostname(r.Spec.Hostname); err != nil {
-		return fmt.Errorf("invalid hostname: %w", err)
-	}
-
-	if err := isValidMachineNetwork(r.Spec.MachineNetwork); err != nil {
-		return fmt.Errorf("invalid machine network: %w", err)
-	}
-
-	return nil
-}
-
 // ValidateUpdate implements webhook.Validator so a webhook will be registered for the type
 func (r *ImageClusterInstall) ValidateUpdate(old runtime.Object) (admission.Warnings, error) {
 	icilog.Info("validate update", "name", r.Name)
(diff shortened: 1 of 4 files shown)


controllers/imageclusterinstall_controller.go
```diff
@@ -227,13 +226,6 @@ func (r *ImageClusterInstallReconciler) Reconcile(ctx context.Context, req ctrl.
 		}
 	}
 
-	if ici.Status.BareMetalHostRef != nil && !v1alpha1.BMHRefsMatch(ici.Spec.BareMetalHostRef, ici.Status.BareMetalHostRef) {
```
tsorya, line 230: why? I am not sure we need it too much but still, the idea is that you can sort of change bmh and will be nice to cleanup the old one but we can drop it i believe

tsorya:
lets split to 2 prs maybe?

api/v1alpha1/imageclusterinstall_webhook.go
```diff
@@ -53,21 +56,6 @@ func (r *ImageClusterInstall) ValidateCreate() (admission.Warnings, error) {
 	return nil, nil
 }
 
-func (r *ImageClusterInstall) validate() error {
```
tsorya, line 56: Should not it run  on create too?

tsorya: sorry, seems like it just changed it place

api/v1alpha1/imageclusterinstall_webhook.go
```diff
@@ -81,20 +69,43 @@ func (r *ImageClusterInstall) ValidateUpdate(old runtime.Object) (admission.Warn
...
+	if err := isValidMachineNetwork(r.Spec.MachineNetwork); err != nil {
+		return fmt.Errorf("invalid machine network: %w", err)
+	}
+	return nil
+}
+
+func installationStarted(r *ImageClusterInstall) bool {
+	// Block update if the image is ready
+	// Consider updating this to check the ClusterInstallCompleted condition
+	imageReadyCondition := findCondition(r.Status.Conditions, hivev1.ClusterInstallRequirementsMet)
```
tsorya, line 105: hmm, i am not sure we update this condition anymore to say the truth, I think better to validate if there is bmh in status or even in reference

api/v1alpha1/imageclusterinstall_webhook.go
```diff
@@ -81,20 +69,43 @@ func (r *ImageClusterInstall) ValidateUpdate(old runtime.Object) (admission.Warn
...
 	return nil, nil
 }
 
+// ValidateDelete implements webhook.Validator so a webhook will be registered for the type
+func (r *ImageClusterInstall) ValidateDelete() (admission.Warnings, error) {
+	return nil, nil
+}
+
+func (r *ImageClusterInstall) validate() error {
+	if err := isValidSSHPublicKey(r.Spec.SSHKey); err != nil {
```
tsorya, line 89: Maybe add validation that we have referenced bmh?

api/v1alpha1/imageclusterinstall_webhook.go
tsorya: Or we can add this condition here https://github.com/openshift/image-based-install-operator/blob/main/controllers/imageclusterinstall_controller.go#L277

api/v1alpha1/imageclusterinstall_webhook.go
eranco74: What for?

tsorya: as we don't support ici without bmh anymore? but it's ok if we allow to set it after

api/v1alpha1/imageclusterinstall_webhook.go
```diff
@@ -142,19 +153,11 @@ func isValidMachineNetwork(machineNetwork string) error {
...
-}
-
-func BMHRefsMatch(ref1 *BareMetalHostReference, ref2 *BareMetalHostReference) bool {
-	if ref1 == nil && ref2 == nil {
-		return true
-	}
-
-	if ref1 == nil && ref2 != nil || ref1 != nil && ref2 == nil {
-		return false
+func findCondition(conditions []hivev1.ClusterInstallCondition, condType hivev1.ClusterInstallConditionType) *hivev1.ClusterInstallCondition {
```
tsorya, line 156: can't you use function from conditions?

api/v1alpha1/imageclusterinstall_webhook_test.go
```diff
@@ -240,43 +273,24 @@ var _ = Describe("ValidateUpdate", func() {
...
+			Message: InstallSucceededMessage,
 		}}
 
 		warns, err := newClusterInstall.ValidateUpdate(oldClusterInstall)
 		Expect(warns).To(BeNil())
 		Expect(err).To(BeNil())
 	})
 
-	It("metadata update succeeds when BMH ref is set", func() {
+	It("metadata update succeeds when iamge is ready", func() {
```
tsorya, line 293: nit: iamge -> image

PR #96, opened 2024-09-12
[backplane-2.7] Update to go 1.22

carbonin:
This requires some special handling over the changes in the main branch because we have konflux running in the release branch. The first three commits are direct cherry-picks of the commits in main, but the last one exists to make the konflux system happy.

Adding a konflux-specific Dockerfile is required because the only go 1.22 image accessible from the konflux build environment is the one in brew.registry.redhat.io, but that image is not available in openshift CI.

Likewise the ci image isn't available in konflux. Until we have an image that satisfies both systems we'll need to maintain two Dockerfiles.


carbonin:
@eliorerz @tsorya I'm hoping https://github.com/openshift/release/pull/56657 will solve the test failures?

carbonin:
Looking better :tada:

PR #100, opened 2024-09-15
MGMT-18899: Reboot externally provisioned node if it was already powered on

eranco74:
Set the ici.Status.BareMetalHostRef only after the BMH provisioning is set by IBIO


eranco74:
Here is the BMH provisioning state machine, that dictated the required UTs:
https://github.com/metal3-io/baremetal-operator/blob/main/docs/baremetalhost-states.md

PR #102, opened 2024-09-15
Reduce log spam

eranco74:
Currently this log line show up 52 times per cluster we install

### DIFF
--- a/controllers/imageclusterinstall_controller.go
+++ b/controllers/imageclusterinstall_controller.go
@@ -330,7 +330,6 @@ func (r *ImageClusterInstallReconciler) validateSeedReconfigurationWithBMH(
 	// Skip validations in case of the state is Externally Provisioned as it will not be inspected
 	if bmh.Spec.ExternallyProvisioned {
 		msg := fmt.Sprintf("BareMetalHost %s/%s, is externally provisioned, skipping hardware validation", bmh.Namespace, bmh.Name)
-		r.Log.Info(msg)
 		if updateErr := r.setRequirementsMetCondition(ctx, ici, corev1.ConditionTrue, v1alpha1.HostValidationSucceeded, msg); updateErr != nil {
 			r.Log.WithError(updateErr).Error("failed to update ImageClusterInstall status")
 		}
@@ -340,7 +339,6 @@ func (r *ImageClusterInstallReconciler) validateSeedReconfigurationWithBMH(
 	// no need to validate if inspect annotation is disabled
 	if bmh.ObjectMeta.Annotations != nil && !isInspectionEnabled(bmh) {
 		msg := fmt.Sprintf("inspection is disabled for BareMetalHost %s/%s, skip hardware validation", bmh.Namespace, bmh.Name)
-		r.Log.Info(msg)
 		if updateErr := r.setRequirementsMetCondition(ctx, ici, corev1.ConditionTrue, v1alpha1.HostValidationSucceeded, msg); updateErr != nil {
 			r.Log.WithError(updateErr).Error("failed to update ImageClusterInstall status")
 		}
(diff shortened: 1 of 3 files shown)


controllers/imageclusterinstall_controller.go
```diff
@@ -340,9 +340,12 @@ func (r *ImageClusterInstallReconciler) validateSeedReconfigurationWithBMH(
 	// no need to validate if inspect annotation is disabled
 	if bmh.ObjectMeta.Annotations != nil && !isInspectionEnabled(bmh) {
 		msg := fmt.Sprintf("inspection is disabled for BareMetalHost %s/%s, skip hardware validation", bmh.Namespace, bmh.Name)
-		r.Log.Info(msg)
-		if updateErr := r.setRequirementsMetCondition(ctx, ici, corev1.ConditionTrue, v1alpha1.HostValidationSucceeded, msg); updateErr != nil {
-			r.Log.WithError(updateErr).Error("failed to update ImageClusterInstall status")
+		condition := findCondition(ici.Status.Conditions, v1alpha1.HostValidationSucceeded)
```
tsorya, line 343: Actually i don't think we need this change, this was relevant to previous code version but not for new one,
we will set bmh.Spec.ExternallyProvisioned right after first run in case !isInspectionEnabled(bmh) so this case is not relevant anymore

eranco74: Where does that happen?
AFAIK, [this](https://github.com/openshift/image-based-install-operator/blob/7200975383c1a07fc97b6f60f4b11737b34ec0b9/controllers/imageclusterinstall_controller.go#L613) is the only place where we set the bmh.Spec.ExternallyProvisioned

eranco74:
e2e failed.
ICI status:
```
  - lastProbeTime: "2024-09-15T15:59:50Z"
    lastTransitionTime: "2024-09-15T15:59:50Z"
    message: "Cluster is installing\nClusterVersion Status: Failed to get ibi-monitor-cm
      : failed to get API group resources: unable to retrieve the complete list of
      server APIs: v1: Get \"https://api.ibi-cluster.mydomain.com:6443/api/v1?timeout=10s\":
      dial tcp 192.168.111.26:6443: connect: connection refused\nNodes Status: "
 
 ```
 From the operator logs, it seems that we have few other spamming log lines:
 1. skipping hardware validation
 2. Crypto keys already exist
 3. Admin password already exists
 4. Kubeconfig already exists

 https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_image-based-install-operator/102/pull-ci-openshift-image-based-install-operator-main-e2e-ibio/1835323657711259648/artifacts/e2e-ibio/image-based-install-operator-gather/artifacts/root/ibio-gather/image-based-install-operator-manager.yaml


Also, note that we had 24 reconciles until we started to monitor the installation :exploding_head: - we need to look into it! 
Overall we reconcile the CR  86 times, the test failed on timeout so I hope that on success it's less

eranco74:
/test e2e-ibio
Unsure why it failed, we need to collect logs from the spoke
Rerunning...

controllers/imageclusterinstall_controller.go
tsorya: Exactly, this code will run only once cause in case externally provisioned was not set we will set it on the first run. Second time we will be here we will exit 
https://github.com/openshift/image-based-install-operator/pull/102/files#diff-2b950e24497af9d913e379e1aad24d9446d374d42e696b0ee2a0bd5079b3b5a2R331

eranco74: We will only set it once the BMH is available which may take some time ( a lot of time if it goes to error).
Also in you link I can see another spamming log message:
```BareMetalHost ... is externally provisioned, skipping hardware validation```
Fixing that as well

tsorya: Good

PR #103, opened 2024-09-15
MGMT-18898: Allow transition from ClusterInstallationTimedOut to installComplete

tsorya:
[MGMT-18898](https://issues.redhat.com//browse/MGMT-18898): Allow transition from ClusterInstallationTimedOut to installComplete
In case installation timed out but reconcilation ran we still want to validate if cluster managed to be installed or not, in case it finished successfully after timeout we will set installed

### DIFF
--- a/controllers/imageclusterinstall_controller.go
+++ b/controllers/imageclusterinstall_controller.go
@@ -145,7 +145,7 @@ func (r *ImageClusterInstallReconciler) Reconcile(ctx context.Context, req ctrl.
 	}
 
 	// Nothing to do if the installation process has already stopped
-	if installationStopped(ici) {
+	if installationCompleted(ici) {
 		log.Infof("Cluster %s/%s finished installation process, nothing to do", ici.Namespace, ici.Name)
 		return ctrl.Result{}, nil
 	}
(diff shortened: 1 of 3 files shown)


controllers/imageclusterinstall_controller.go
```diff
@@ -436,24 +430,42 @@ func (r *ImageClusterInstallReconciler) checkClusterTimeout(ctx context.Context,
...
-			log.WithError(err).Error("failed to set installing conditions")
+			r.Log.WithError(err).Error("failed to set installing conditions")
 		}
 		return ctrl.Result{RequeueAfter: time.Minute}, nil
 	}
-	log.Info("cluster is installed")
+	r.Log.Info("cluster is installed")
+
+	// After installation ended we don't want that ironic will do any changes in the node
+	patch := client.MergeFrom(bmh.DeepCopy())
```
tsorya, line 459: was move here in order to set installed only after setting detached, in case setting detached will fail we will reconcile one more time

eranco74 (commented):
/lgtm

carbonin:
Based on the definitions in https://github.com/openshift/hive/blob/master/docs/enhancements/cluster-install-apis.md I think we're not using these conditions the correct way as of this change.

Specifically "stopped".

From the doc:
>True once the controller is no longer working to reach desired state. Does not imply success or failure by itself, but can be combined with Completed/Failed.

If we're actually continuing to monitor the cluster, why not just not do a timeout or extend it?

PR #107, opened 2024-09-17
MGMT-18944: Remove route

tsorya:


carbonin:
Probably will need to add the route in manually for the integration test.

Unless we just don't want to download the image anymore :shrug:

hack/integration-test/validate.sh
```diff
@@ -16,6 +16,3 @@ if [[ -z "$url" ]]; then
   echo "ERROR: configurationImageURL on ImageClusterInstall ibi-test/ibi-test was not set within 60 seconds"
```
carbonin, line 16: We're not really checking this any more. Can we update these messages to reflect that we're checking the data image not the ICI?

hack/integration-test/validate.sh
```diff
@@ -16,6 +16,3 @@ if [[ -z "$url" ]]; then
...
   exit 1
 fi
-
-curl --insecure --output "test.iso" "$url"
-file "test.iso" | grep -q "ISO 9660 CD-ROM"
```
carbonin, line 21: I guess this is fine to remove now that we have an e2e test :shrug:

hack/integration-test/validate.sh
carbonin: This doesn't really apply to this change so we can do this in some other PR

carbonin (approved):

PR #110, opened 2024-09-17
NO-ISSUE: Use logger with fields everywhere

carbonin:
Any time r.Log is used the ImageClusterInstall name and namespace are not being logged with the message which makes them much less useful.

### DIFF
--- a/controllers/imageclusterinstall_controller.go
+++ b/controllers/imageclusterinstall_controller.go
@@ -19,6 +19,7 @@ package controllers
 import (
 	"bytes"
 	"context"
+
 	// These are required for image parsing to work correctly with digest-based pull specs
 	// See: https://github.com/opencontainers/go-digest/blob/v1.0.0/README.md#usage
 	_ "crypto/sha256"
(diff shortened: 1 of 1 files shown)


2uasimojo (commented):
Missed one: https://github.com/openshift/image-based-install-operator/blob/d4befdf99ba0653b5a9d72745ed9aea4335d7fe1/controllers/imageclusterinstall_controller.go#L554

controllers/imageclusterinstall_controller.go
```diff
@@ -259,7 +260,7 @@ func (r *ImageClusterInstallReconciler) Reconcile(ctx context.Context, req ctrl.
 	}
 	if !annotationExists(&bmh.ObjectMeta, ibioManagedBMH) {
 		// TODO: consider replacing this condition with `dataDisk.Status.AttachedImage`
-		r.Log.Info("Nothing to do, waiting for BMH to get %s annotation", ibioManagedBMH)
+		log.Info("Nothing to do, waiting for BMH to get %s annotation", ibioManagedBMH)
```
2uasimojo, line 263:
```suggestion
		log.Infof("Nothing to do, waiting for BMH to get %s annotation", ibioManagedBMH)
```

carbonin:
>Missed one:

This one is actually outside of the reconcile loop so it's not specific to a particular resource.

PR #111, opened 2024-09-18
MGMT-18909: Delete DataImage on ICI deletion

eliorerz:
/cc @eliorerz

### DIFF
--- a/controllers/imageclusterinstall_controller.go
+++ b/controllers/imageclusterinstall_controller.go
@@ -586,9 +586,9 @@ func (r *ImageClusterInstallReconciler) SetupWithManager(mgr ctrl.Manager) error
 		Complete(r)
 }
 
-func (r *ImageClusterInstallReconciler) getBmhDataImage(ctx context.Context, bmh *bmh_v1alpha1.BareMetalHost) *bmh_v1alpha1.DataImage {
+func (r *ImageClusterInstallReconciler) getBmhDataImage(ctx context.Context, log logrus.FieldLogger, bmh *bmh_v1alpha1.BareMetalHost) (*bmh_v1alpha1.DataImage, error) {
 	if bmh == nil {
-		return nil
+		return nil, nil
 	}
 	dataImage := bmh_v1alpha1.DataImage{}
 	key := client.ObjectKey{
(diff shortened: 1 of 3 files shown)


controllers/imageclusterinstall_controller.go
```diff
@@ -586,21 +586,25 @@ func (r *ImageClusterInstallReconciler) SetupWithManager(mgr ctrl.Manager) error
...
 	key := client.ObjectKey{
 		Name:      bmh.Name,
 		Namespace: bmh.Namespace,
 	}
 	err := r.Get(ctx, key, &dataImage)
-	if apierrors.IsNotFound(err) {
-		return nil
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			r.Log.Warnf("Can't find DataImage %s/%s", bmh.Namespace, bmh.Name)
```
tsorya, line 601: pass log to the function as it will have withfields logic

controllers/imageclusterinstall_controller.go
```diff
@@ -711,8 +718,19 @@ func (r *ImageClusterInstallReconciler) removeBMHDataImage(ctx context.Context,
 
 	if bmh != nil {
```
tsorya, line 719:
```suggestion
	if bmh != nil {
```
```suggestion
	if bmh == nil {
	return
	}
```

eliorerz: Why? BMH shouldn't be nil if we want to change/remove some of its annotations

controllers/imageclusterinstall_controller.go
```diff
@@ -708,14 +715,26 @@ func (r *ImageClusterInstallReconciler) removeBMHDataImage(ctx context.Context,
 	if err := r.Client.Delete(ctx, dataImage); err != nil {
 		return err
 	}
+	if bmh == nil {
+		return nil
```
eranco74, line 719: This deserves a log, it means we can never delete this dataimage

controllers/imageclusterinstall_controller.go
```diff
@@ -586,21 +586,25 @@ func (r *ImageClusterInstallReconciler) SetupWithManager(mgr ctrl.Manager) error
 		Complete(r)
 }
 
-func (r *ImageClusterInstallReconciler) getBmhDataImage(ctx context.Context, bmh *bmh_v1alpha1.BareMetalHost) *bmh_v1alpha1.DataImage {
+func (r *ImageClusterInstallReconciler) getBmhDataImage(ctx context.Context, log logrus.FieldLogger, bmh *bmh_v1alpha1.BareMetalHost) (*bmh_v1alpha1.DataImage, error) {
 	if bmh == nil {
-		return nil
+		return nil, nil
```
eranco74, line 591: We don't care for the BMH, we just need the name and namespace.

PR #113, opened 2024-09-18
MGMT-18950: Set condition stopped to false in case of timeout

tsorya:
[MGMT-18950](https://issues.redhat.com//browse/MGMT-18950): Set condition stopped to false in case of timeout
As we allow to update status from timeout to installed we should not set stopped installation flag to false on timeout

### DIFF
--- a/api/v1alpha1/imageclusterinstall_types.go
+++ b/api/v1alpha1/imageclusterinstall_types.go
@@ -34,7 +34,7 @@ const (
 	HostConfigurationSucceededMessage = "Configuration image is attached to the referenced host"
 
 	InstallTimedoutReason  = "ClusterInstallationTimedOut"
-	InstallTimedoutMessage = "Cluster installation has timed out"
+	InstallTimedoutMessage = "Cluster installation is taking longer than expected"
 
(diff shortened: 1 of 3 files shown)


api/v1alpha1/imageclusterinstall_types.go
```diff
@@ -34,7 +34,7 @@ const (
 	HostConfigurationSucceededMessage = "Configuration image is attached to the referenced host"
 
 	InstallTimedoutReason  = "ClusterInstallationTimedOut"
-	InstallTimedoutMessage = "Cluster installation has timed out"
+	InstallTimedoutMessage = "Cluster installation takes longer than expected"
```
carbonin, line 37:
```suggestion
	InstallTimedoutMessage = "Cluster installation is taking longer than expected"
```

carbonin (changes requested):
Looks like `setClusterTimeoutConditions` is still setting `ClusterInstallFailed` to `True`.

I don't think it's valid to set that without also setting `Stopped`. We should probably also change that one.

carbonin (approved):

PR #117, opened 2024-09-22
Cleanup & minor refactor

eranco74:
* Since we no longer set the image URL remove ConfigurationImageURL from ICI status
* Disable AutomatedCleaningMode earlier
* Add const for a soft reboot annotation value
* Set RequirementsMet condition in case of missing clusterDeploymentRef
* Initialize conditions prior to other logic

### DIFF
--- a/api/v1alpha1/imageclusterinstall_types.go
+++ b/api/v1alpha1/imageclusterinstall_types.go
@@ -113,9 +113,6 @@ type ImageClusterInstallStatus struct {
 
 	BareMetalHostRef *BareMetalHostReference `json:"bareMetalHostRef,omitempty"`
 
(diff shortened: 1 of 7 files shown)


eranco74:
/retest
```
ERRO[2024-09-22T09:29:21Z] 
  * could not initialize namespace: could not delete egress firewall: error finding the scope of the object: failed to get restmapping: no matches for kind "EgressFirewall" in group "k8s.ovn.org" 
  ```

carbonin:
@eranco74 is this something we expect to be fixed with a retest or is there something else we need to do?

eranco74:
> @eranco74 is this something we expect to be fixed with a retest or is there something else we need to do?

https://github.com/openshift/ci-tools/pull/4342

PR #120, opened 2024-09-25
MGMT-19012: The controller shouldn't remove the finalizer as long the dataImage exists

eranco74:
The finalizer will be removed by the controller if the referenced BareMetalHost or the DataImage no longer exists.

### DIFF
--- a/controllers/imageclusterinstall_controller.go
+++ b/controllers/imageclusterinstall_controller.go
@@ -697,33 +697,27 @@ func (r *ImageClusterInstallReconciler) getBMH(ctx context.Context, bmhRef *v1al
 	return bmh, nil
 }
 
-func (r *ImageClusterInstallReconciler) removeBMHDataImage(ctx context.Context, log logrus.FieldLogger, bmh *bmh_v1alpha1.BareMetalHost, bmhRef *v1alpha1.BareMetalHostReference) error {
-	dataImage := &bmh_v1alpha1.DataImage{}
-
-	key := types.NamespacedName{
-		Name:      bmhRef.Name,
-		Namespace: bmhRef.Namespace,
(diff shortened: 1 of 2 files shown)


controllers/imageclusterinstall_controller.go
```diff
@@ -697,33 +697,27 @@ func (r *ImageClusterInstallReconciler) getBMH(ctx context.Context, bmhRef *v1al
...
 
-func (r *ImageClusterInstallReconciler) removeBMHDataImage(ctx context.Context, log logrus.FieldLogger, bmh *bmh_v1alpha1.BareMetalHost, bmhRef *v1alpha1.BareMetalHostReference) error {
-	dataImage := &bmh_v1alpha1.DataImage{}
-
-	key := types.NamespacedName{
-		Name:      bmhRef.Name,
-		Namespace: bmhRef.Namespace,
+func (r *ImageClusterInstallReconciler) removeBMHDataImage(ctx context.Context, log logrus.FieldLogger, bmhRef types.NamespacedName) (*bmh_v1alpha1.DataImage, error) {
+	dataImage, err := r.deleteDataImage(ctx, log, bmhRef)
+	if err != nil || dataImage == nil {
```
carbonin, line 702: If the data image was already removed (`err == nil && dataImage == nil`) should we still attach and reboot the BMH? Looks like we're exiting early in that case here.

eranco74:
/hold
I'd like to run the e2e with [this](https://github.com/openshift/release/pull/57011) change

controllers/imageclusterinstall_controller.go
eranco74: The only reason we attach and reboot is to allow the dataImage to be deleted.
If the dataImage doesn't exist we don't want to touch the BMH.

carbonin: Ah okay

carbonin (approved):
Looks good to me, unhold when you're ready @eranco74

PR #124, opened 2024-10-01
MGMT-19034 IBIO shouldn't requeue in case BMH hasn't started yet

eranco74:
* Reconcile will be triggered once the BMH get updated
* Fixed logging format
* Improve mapBMHToICI using list options to filter the relevant ICI
* Remove useless warning logging saying "Can't find DataImage"
 * Delete test that validate the behavior in case of spec changes after the installation started
    Spec update during installation isn't supported and bllocked by an addmission webhook hence no need to cover this in a test

### DIFF
--- a/controllers/conditions.go
+++ b/controllers/conditions.go
@@ -156,7 +156,7 @@ func (r *ImageClusterInstallReconciler) setRequirementsMetCondition(ctx context.
 	if updated := setClusterInstallCondition(&ici.Status.Conditions, cond); !updated {
 		return nil
 	}
-	r.Log.Info("Setting requirements met condition, status: %s, reason: %s, message: %s", cond.Status, cond.Reason, cond.Message)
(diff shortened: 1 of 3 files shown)


eranco74:
/hold
need to add the bmhRef indexing in the UT

controllers/conditions.go
```diff
@@ -156,7 +156,7 @@ func (r *ImageClusterInstallReconciler) setRequirementsMetCondition(ctx context.
 	if updated := setClusterInstallCondition(&ici.Status.Conditions, cond); !updated {
 		return nil
 	}
-	r.Log.Info("Setting requirements met condition, status: %s, reason: %s, message: %s", cond.Status, cond.Reason, cond.Message)
+	r.Log.Infof("Setting requirements met condition, status: %s, reason: %s, message: %s", cond.Status, cond.Reason, cond.Message)
```
tsorya, line 159: actually we should provide log from outside in order to have fields

PR #127, opened 2024-10-06
MGMT-19033: IBIO OOM while provisioning many clusters

tsorya:
1. Updating memory limit to 1500miB
2. Adding flag for golang garbage collection in order to free memory faster


eranco74 (commented):
/lgtm

PR #129, opened 2024-10-07
MGMT-19041: Duration of cluster installs grows as many clusters are installed concurrently

eranco74:
Move the installation monitor to a separate controller
This change ensures that installation monitoring (which can take a few seconds) doesn't block the reconciliation of ICI CRs that haven't started the installation process.


controllers/imageclusterinstall_controller.go
```diff
@@ -143,9 +138,8 @@ func (r *ImageClusterInstallReconciler) Reconcile(ctx context.Context, req ctrl.
 		return res, err
 	}
 
-	// Nothing to do if the installation process has already stopped
-	if installationCompleted(ici) {
-		log.Infof("Cluster %s/%s finished installation process, nothing to do", ici.Namespace, ici.Name)
+	// Nothing to do if the installation process started
+	if !ici.Status.BootTime.IsZero() {
 		return ctrl.Result{}, nil
```
tsorya, line 143: maybe move it before 
https://github.com/openshift/image-based-install-operator/pull/129/files#diff-2b950e24497af9d913e379e1aad24d9446d374d42e696b0ee2a0bd5079b3b5a2L131 ?

tsorya: hmm, you can't cause of finalizer but i believe there is no point to add reconcile log in case we are not actually?

tsorya: Lets add same logic to mapping functions? If we know we should not reconcile we can filter it earlier, at least for bmh and cd events, will make our queue smaller

controllers/imageclusterinstall_monitor.go
```diff
@@ -0,0 +1,242 @@
...
+// ImageClusterInstallReconciler reconciles a ImageClusterInstall object
+type ImageClusterInstallMonitor struct {
+	client.Client
+	Log                          logrus.FieldLogger
+	Scheme                       *runtime.Scheme
+	DefaultInstallTimeout        time.Duration
+	GetSpokeClusterInstallStatus monitor.GetInstallStatusFunc
+}
+
+//+kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch;create;update;patch;delete
```
tsorya, line 52: why we need those here? seems like just duplication of the same resources

controllers/imageclusterinstall_controller.go
eranco74: There's no mapping for the ICI itself.
Everything that run prior to this needs to happen regardless if the installation started or not.

eranco74:
close in favor of https://github.com/openshift/image-based-install-operator/pull/132

PR #131, opened 2024-10-07
MGMT-18973 IBIO serviceaccount does not have proper permissions to patch extraManifests configmap resources

eranco74:
Allow IBIO to patch configMaps in all namespaces


carbonin:
Isn't this the default? If not we would need the same notation for all the resources we watch.

I was pretty sure [my PR](https://github.com/openshift/image-based-install-operator/pull/115) already fixed this.

PR #139, opened 2024-10-15
OCPBUGS-43330: BMO never mounted dataimage into IBI provisioned SNO

eranco74:
There's a race between the DataImage creation and the reboot annotation added in updateBMHProvisioningState This PR adds a call to get the dataImage with nonCachedClient before adding the reboot annotation. It should create some delay that will hopefully resolve this race.


eranco74:
This isn't good enough

PR #140, opened 2024-10-15
MGMT-19154: BMO never mounted dataimage into IBI provisioned SNO

eranco74:
There's a race between the DataImage creation and the reboot annotation added in updateBMHProvisioningState
This PR adds a check that the DataImage creationTimestamp is more than the `DataImageCoolDownPeriod` ago (set to 1 second by default).
In case the the dataImage was created withing the `DataImageCoolDownPeriod` the controller will
return early and requeue the ICI CR after the `DataImageCoolDownPeriod`.
This should prevent IBIO from adding the reboot annotation before BMO cached client notice that a DataImage exists.

### DIFF
--- a/controllers/imageclusterinstall_controller.go
+++ b/controllers/imageclusterinstall_controller.go
@@ -65,11 +65,12 @@ import (
 )
 
 type ImageClusterInstallReconcilerOptions struct {
-	ServiceName      string `envconfig:"SERVICE_NAME"`
-	ServiceNamespace string `envconfig:"SERVICE_NAMESPACE"`
-	ServicePort      string `envconfig:"SERVICE_PORT"`
-	ServiceScheme    string `envconfig:"SERVICE_SCHEME"`
-	DataDir          string `envconfig:"DATA_DIR" default:"/data"`
+	ServiceName             string        `envconfig:"SERVICE_NAME"`
+	ServiceNamespace        string        `envconfig:"SERVICE_NAMESPACE"`
+	ServicePort             string        `envconfig:"SERVICE_PORT"`
+	ServiceScheme           string        `envconfig:"SERVICE_SCHEME"`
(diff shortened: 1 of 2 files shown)


controllers/imageclusterinstall_controller.go
```diff
@@ -511,18 +517,20 @@ func (r *ImageClusterInstallReconciler) updateBMHProvisioningState(ctx context.C
 	return nil
 }
 
-func (r *ImageClusterInstallReconciler) createBMHDataImage(ctx context.Context, log logrus.FieldLogger, bmh *bmh_v1alpha1.BareMetalHost, url string) error {
-	_, err := r.getDataImage(ctx, bmh.Namespace, bmh.Name)
+// createBMHDataImage will create a dataImage with the URL for the config ISO if dataImage didn't exist
```
tsorya, line 520: i think we need to change a comment

carbonin:
Is this supposed to resolve that bug? If so should we move it to MGMT or ACM and file it against IBIO properly?

Or are we waiting for some fix from the BMO team? In which case we should open a separate issue to mitigate this until it is fixed on their side.

eranco74:
> Is this supposed to resolve that bug? If so should we move it to MGMT or ACM and file it against IBIO properly?
> 
> Or are we waiting for some fix from the BMO team? In which case we should open a separate issue to mitigate this until it is fixed on their side.

This should resolve the bug, it's still unknown if BMO can going to make a proper fix for it. 
So I guess let's open another one for MGMT and link it to OCPBUGS-43330

carbonin:
/hold

No more changes to 2.7 as we're in RC now

eranco74:
> /hold
> 
> No more changes to 2.7 as we're in RC now

Can we push this already?

carbonin:
Not quite yet. Release is still not out

PR #141, opened 2024-10-16
MGMT-19188: BMO never mounted dataimage into IBI provisioned SNO

tsorya:
There's a race between the DataImage creation and the reboot annotation added in updateBMHProvisioningState This PR adds a check that the DataImage creationTimestamp is more than the DataImageCoolDownPeriod ago (set to 1 second by default). In case the the dataImage was created withing the DataImageCoolDownPeriod the controller will return early and requeue the ICI CR after the DataImageCoolDownPeriod. This should prevent IBIO from adding the reboot annotation before BMO cached client notice that a DataImage exists.

### DIFF
--- a/controllers/imageclusterinstall_controller.go
+++ b/controllers/imageclusterinstall_controller.go
@@ -34,9 +34,8 @@ import (
 	"strings"
 	"time"
 
-	"github.com/google/uuid"
-
 	"github.com/containers/image/v5/docker/reference"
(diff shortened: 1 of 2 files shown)


carbonin:
What's the difference between this and https://github.com/openshift/image-based-install-operator/pull/140?

eranco74:
> What's the difference between this and #140?

This is for main

PR #145, opened 2024-10-31
MGMT-19203: Update the reason in case of timeout while bmh is powered off

giladravid16:
Currently if we reach the cluster installation timeout while the bmh is powered off, then we always set the reason to "ClusterInstallationInProgress" and the message to "Waiting for BMH to power on".
This pr adds a check to see if we reached the timeout, and if so sets the conditions accordingly.

closes  [MGMT-19203](https://issues.redhat.com//browse/MGMT-19203)

### DIFF
--- a/controllers/imageclusterinstall_monitor.go
+++ b/controllers/imageclusterinstall_monitor.go
@@ -93,6 +93,15 @@ func (r *ImageClusterInstallMonitor) monitorInstallationProgress(
 	}
 	if !bmh.Status.PoweredOn {
 		log.Infof("BareMetalHost %s/%s is not powered on yet", bmh.Name, bmh.Namespace)
+		timedout, err := r.handleClusterTimeout(ctx, log, ici, r.DefaultInstallTimeout)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+		if timedout {
+			log.Infof("BareMetalHost %s/%s failed to power on within the cluster installation timeout", bmh.Name, bmh.Namespace)
+			// in case of timeout we want to requeue after 1 hour
+			return ctrl.Result{RequeueAfter: time.Hour}, nil
+		}
 		if err := r.setClusterInstallingConditions(ctx, ici, "Waiting for BMH to power on"); err != nil {
 			log.WithError(err).Error("failed to set installing conditions")
 		}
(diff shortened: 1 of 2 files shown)


controllers/imageclusterinstall_monitor.go
```diff
@@ -93,6 +93,14 @@ func (r *ImageClusterInstallMonitor) monitorInstallationProgress(
 	}
 	if !bmh.Status.PoweredOn {
 		log.Infof("BareMetalHost %s/%s is not powered on yet", bmh.Name, bmh.Namespace)
+		timedout, err := r.handleClusterTimeout(ctx, log, ici, r.DefaultInstallTimeout, "BMH failed to power on within the cluster installation timeout (%s)")
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+		if timedout {
+			// in case of timeout we want to requeue after 1 hour
+			return ctrl.Result{RequeueAfter: time.Hour}, nil
```
tsorya, line 102: @eranco74 we want to requeu after an hour? I don't see it in ticket, as for it should be fine

giladravid16: In case of a timeout during the installation, we [requeue after an hour](https://github.com/openshift/image-based-install-operator/blob/main/controllers/imageclusterinstall_monitor.go#L150-L153). I kept it an hour here as well because I didn't see why there should be a different between the 2 cases.

tsorya: I see, as we are not watching bmh in this controller, seems logical to do it. Thanks

carbonin: Should we maybe log here? At least debug?

eranco74: I think we can move the code added here outside of the `if !bmh.Status.PoweredOn` scope (move it to line 94).
The timeout condition message doesn't need to specify the BMH condition (The BMH status will be reflected in the installing condition message) .
This should allow removing the call to`handleClusterTimeout` from `checkClusterStatus` .
Thoughts?

eranco74: NM, if we do that we can't recover from timeout.
So just add the log (info is fine).
Note that the the timeout condition message can be the same regardless of the BMH status

giladravid16: I added the log and reverted my changes to the message so it will be the same

PR #146, opened 2024-10-31
MGMT-19107: Make MachineNetwork required when setting Proxy

giladravid16:
Installation always fails when MachineNetwork is not set and Proxy is set, so it's best to reject the creation of the ICI in the first place.

closes [MGMT-19107](https://issues.redhat.com//browse/MGMT-19107)

### DIFF
--- a/api/v1alpha1/imageclusterinstall_webhook.go
+++ b/api/v1alpha1/imageclusterinstall_webhook.go
@@ -93,6 +93,9 @@ func (r *ImageClusterInstall) validate() error {
 	if err := isValidMachineNetwork(r.Spec.MachineNetwork); err != nil {
 		return fmt.Errorf("invalid machine network: %w", err)
 	}
+	if err := isValidProxy(r.Spec.Proxy, r.Spec.MachineNetwork); err != nil {
+		return fmt.Errorf("invalid proxy: %w", err)
+	}
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


api/v1alpha1/imageclusterinstall_webhook.go
```diff
@@ -147,3 +150,10 @@ func isValidMachineNetwork(machineNetwork string) error {
 	}
 	return nil
 }
+
+func isValidNoProxy(proxy *Proxy, machineNetwork string) error {
```
tsorya, line 154: I think there is small misunderstanding, it should be relevant to any proxy field and not for no proxy specifically.
More of it , no proxy field is the less relevant here, we should check if http or https proxy was set

giladravid16: [This](https://issues.redhat.com/browse/MGMT-19107?focusedId=25926359&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-25926359) comment made by Nick, made me think the issue was only with no_proxy. If that's not the case I'll change it.

tsorya: The idea was that we need machine network for creating right no proxy but no proxy is do nothing by itself.
We need to verify http and https proxy params as they actually set proxy

tsorya: Nope. The real issue is that if http or https proxy is defined we require machine network

carbonin (approved):

PR #147, opened 2024-10-31
MGMT-19206: installer should save seed reconfiguration file when creating the iso

tsorya:
[MGMT-19206](https://issues.redhat.com//browse/MGMT-19206): installer should save seed reconfiguration file when creating the iso
It will be saved as inside working folder as cluster-configuration/manifest.json
New secret will be created with it's data

### DIFF
--- a/controllers/imageclusterinstall_controller.go
+++ b/controllers/imageclusterinstall_controller.go
@@ -831,6 +831,11 @@ func (r *ImageClusterInstallReconciler) ensureCreds(ctx context.Context, log log
 	if err := r.Credentials.EnsureKubeconfigSecret(ctx, log, cd, filepath.Join(workDir, authDir, credentials.Kubeconfig)); err != nil {
 		return fmt.Errorf("failed to ensure kubeconfig secret: %w", err)
 	}
+
+	if err := r.Credentials.EnsureSeedReconfigurationSecret(ctx, log, cd, filepath.Join(workDir, ClusterConfigDir, credentials.SeedReconfigurationFileName)); err != nil {
+		return fmt.Errorf("failed to ensure seed reconfiguration secret %w", err)
+	}
+
 	return nil
 }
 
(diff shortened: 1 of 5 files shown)


carbonin:
@tsorya when you get back, let me know if this fully addresses https://issues.redhat.com/browse/MGMT-19206

tsorya:
> @tsorya when you get back, let me know if this fully addresses https://issues.redhat.com/browse/MGMT-19206

it will leave cluster-configuration/manifest.json in working dir after creating iso, i can push it new secret or it can be part of  another PR.

carbonin:
We can leave that for another PR, I created that issue just to make sure what we need was accessible.

carbonin:
I think this might actually handle https://issues.redhat.com/browse/MGMT-19207, right?

internal/credentials/credentials.go
```diff
@@ -159,6 +186,10 @@ func KubeadminPasswordSecretName(clusterDeploymentName string) string {
 	return clusterDeploymentName + "-admin-password"
 }
 
+func SeedReconfigurationSecretName(clusterDeploymentName string) string {
```
carbonin, line 189: I don't think this needs to be public.

internal/credentials/credentials.go
```diff
@@ -27,6 +27,7 @@ const (
 	Kubeconfig          = "kubeconfig"
 	kubeadmincreds      = "kubeadmincreds"
 	kubeAdminKey        = "password"
+	seedReconfiguration = "seedReconfiguration"
```
carbonin, line 30: Is there any convention for key names in secrets? I feel like camel case is uncommon. Maybe `seed-reconfiguration` or `manifest.json` since that's the actual file name?

internal/credentials/credentials.go
tsorya: It is used in the controller test so for not copy pasting purposing, same for other functions that create secret names

carbonin (approved):

controllers/imageclusterinstall_controller.go
```diff
@@ -831,6 +831,11 @@ func (r *ImageClusterInstallReconciler) ensureCreds(ctx context.Context, log log
 	if err := r.Credentials.EnsureKubeconfigSecret(ctx, log, cd, filepath.Join(workDir, authDir, credentials.Kubeconfig)); err != nil {
 		return fmt.Errorf("failed to ensure kubeconfig secret: %w", err)
 	}
+
+	if err := r.Credentials.EnsureSeedReconfigurationSecret(ctx, log, cd, filepath.Join(workDir, ClusterConfigDir, credentials.SeedReconfigurationFileName)); err != nil {
```
carbonin, line 835: Why does this file path have `ClusterConfigDir` in it twice?

`workDir` is already `r.Options.DataDir/namespaces/ici.Namespace/string(ici.ObjectMeta.UID)/files/ClusterConfigDir`

So now we end up with this path being `/data/namespaces/ibi-test/377e0bc8-939f-40ac-859a-826853bbded3/files/cluster-configuration/cluster-configuration/manifest.json`

Was that intentional?

PR #154, opened 2024-11-15
[WIP] MGMT-19208: Use installer data from reconfig secret if present

carbonin:
In order to support IBBF IBIO needs to be able reuse the previously generated data when installing a "new" cluster.

This PR brings forward the crypto data as well as the cluster and infra IDs by reading that data from a saved secret if it exists.

For the IDs, this is done where we would previously have generated new IDs and saved them to the cluster metadata on the ImageClusterInstall. This is then passed down to the installer and used as input when generating the image based install config.

For the crypto data, we split the call to create the ISO into two parts. We now generate new manifest data, then overwrite the crypto fields in the manifest on the filesystem, then generate the final ISO which will now have both the correct IDs and crypto data.

Looking for an initial review, but leaving this as WIP since I still haven't actually tested it live.

### DIFF
--- a/controllers/imageclusterinstall_controller.go
+++ b/controllers/imageclusterinstall_controller.go
@@ -765,6 +765,10 @@ func (r *ImageClusterInstallReconciler) writeInputData(
 			return fmt.Errorf("failed to create installation manifest: %w", err)
 		}
 
+		if err := r.Credentials.ImportSeedReconfigurationCrypto(ctx, log, cd, filepath.Join(clusterConfigPath, ClusterConfigDir, credentials.SeedReconfigurationFileName)); err != nil {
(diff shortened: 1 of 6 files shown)


carbonin:
Closing this as this approach won't actually work.

The initial run to generate the base reconfig structure will create new auth an a state file for the installer which will end up being used instead of our crypto.

It would be better to create the first seed reconfig using a temp work dir (ideally just in memory), overlay our crypto from the secret onto that, then write just what we need into the workdir (which won't include the image based, and install configs).

Because this is significantly different from the initial install flow I'm going to close this and make them completely different functions based on the existence of the crypto secrets (kubeconfig, kubeadminpassword, and reconfig).

PR #160, opened 2024-11-27
MGMT-19375: don't reboot the node if data image was mounted already

tsorya:
Sometimes if bmh has externally-provisioned and only true there is a chance that we will create data image before bmh was started so after creating it it will be mounted without a need to reboot the node. Though we will still set reboot node flag and in case machine boots quickly we can get reboot in the middle of the installation.

We should verify that data image was not mounted before setting reboot flag

### DIFF
--- a/controllers/imageclusterinstall_controller.go
+++ b/controllers/imageclusterinstall_controller.go
@@ -257,7 +257,7 @@ func (r *ImageClusterInstallReconciler) Reconcile(ctx context.Context, req ctrl.
 		// notified about the newly created DataImage before adding the reboot annotation in updateBMHProvisioningState
 		return ctrl.Result{RequeueAfter: r.Options.DataImageCoolDownPeriod}, err
 	}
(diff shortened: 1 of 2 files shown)


controllers/imageclusterinstall_controller.go
```diff
@@ -508,7 +508,7 @@ func (r *ImageClusterInstallReconciler) updateBMHProvisioningState(ctx context.C
 		bmh.Spec.Online = true
 		log.Infof("Setting BareMetalHost (%s/%s) spec.Online to true", bmh.Namespace, bmh.Name)
 	}
-	if setAnnotationIfNotExists(&bmh.ObjectMeta, rebootAnnotation, rebootAnnotationValue) {
+	if dataImage.Status.AttachedImage.URL == "" && !setAnnotationIfNotExists(&bmh.ObjectMeta, rebootAnnotation, rebootAnnotationValue) {
```
eranco74, line 511: Should we get a fresh copy of the data image here?

PR #161, opened 2024-11-28
MGMT-19268: IBIO stops serving dataimages to clusters

eranco74:
Solves: IBIO stops serving dataimages to clusters deploying - failed to create iso - error walking tree: could not get pwd: getwd: no such file or directory

Although the image server is using different `workdirs` and different `outPath`, there seems to be a race condition in the create ISO process that results in a corrupted filesystem and missing/inconsistent files or directories. This change ensures that the filesystem creation and finalization operations are not executed concurrently across multiple goroutines, preventing potential data corruption.

### DIFF
--- a/internal/imageserver/imageserver.go
+++ b/internal/imageserver/imageserver.go
@@ -9,6 +9,7 @@ import (
 	"path/filepath"
 	"regexp"
 	"strings"
+	"sync"
 
 	"github.com/diskfs/go-diskfs"
 	"github.com/diskfs/go-diskfs/disk"
@@ -23,6 +24,7 @@ type Handler struct {
 	Log        logrus.FieldLogger
 	WorkDir    string
 	ConfigsDir string
+	Mu         sync.Mutex
 }
 
 var pathRegexp = regexp.MustCompile(`^/images/(.+)/(.+)\.iso$`)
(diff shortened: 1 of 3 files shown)


internal/imageserver/imageserver.go
```diff
@@ -53,7 +55,9 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	// if anything fails remove the workdir, if create succeeds it will remove the workdir so this will be a noop
-	defer os.RemoveAll(isoWorkDir)
+	defer func() {
```
tsorya, line 58: i think it was left this way by mistake, not sure we need a change here

internal/imageserver/imageserver.go
```diff
@@ -74,20 +78,21 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
-
 	outPath, err := tempFileName(h.WorkDir)
 	if err != nil {
 		h.Log.WithError(err).Error("failed to create iso output file")
 		w.WriteHeader(http.StatusInternalServerError)
 		return
 	}
-	if err := create(outPath, isoWorkDir, lca_api.BlockDeviceLabel); err != nil {
-		h.Log.WithError(err).Error("failed to create iso")
+	defer func() {
```
tsorya, line 87:
```suggestion
	defer os.Remove(outPath)
```

PR #164, opened 2024-11-28
MGMT-19341: IBIO isn't creating the config after restart for ICI with bootTime set

eranco74:
Make sure the clusterInfo file exists in case the bootTime is set
This would ensure the clusterInfo is regenerated in case the IBIO pod was restarted

### DIFF
--- a/controllers/imageclusterinstall_controller.go
+++ b/controllers/imageclusterinstall_controller.go
@@ -34,11 +34,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/containers/image/v5/docker/reference"
-	"github.com/google/uuid"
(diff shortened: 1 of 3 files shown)


controllers/imageclusterinstall_controller.go
```diff
@@ -1149,6 +1156,18 @@ func (r *ImageClusterInstallReconciler) writeIBIOStartTimeCM(filePath string) er
 	return nil
 }
 
+func (r *ImageClusterInstallReconciler) clusterInfoFileExists(ici *v1alpha1.ImageClusterInstall) (bool, error) {
```
tsorya, line 1159: hmm, probably in main we should check if iso exists and not manifest json

PR #165, opened 2024-12-01
MGMT-19268: IBIO stops serving dataimages to clusters deploying - fai…

eranco74:
…led to create iso - error walking tree: could not get pwd: getwd: no such file or directory

Although the image server is using diffrent workdirs and diffrent outPath there seems to be a race condition in the create ISO that results a currpteed filesystem and missing/inconsistent files or directories. This change ensures that the filesystem creation and finalization operations are not executed concurrently across multiple goroutines, preventing potential data corruption.

### DIFF
--- a/internal/imageserver/imageserver.go
+++ b/internal/imageserver/imageserver.go
@@ -9,6 +9,7 @@ import (
 	"path/filepath"
 	"regexp"
 	"strings"
+	"sync"
 
 	"github.com/diskfs/go-diskfs"
 	"github.com/diskfs/go-diskfs/disk"
@@ -23,6 +24,7 @@ type Handler struct {
 	Log        logrus.FieldLogger
 	WorkDir    string
 	ConfigsDir string
+	Mu         sync.Mutex
 }
 
 var pathRegexp = regexp.MustCompile(`^/images/(.+)/(.+)\.iso$`)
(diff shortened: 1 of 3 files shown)


internal/imageserver/imageserver.go
```diff
@@ -74,20 +76,19 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
-
 	outPath, err := tempFileName(h.WorkDir)
 	if err != nil {
 		h.Log.WithError(err).Error("failed to create iso output file")
 		w.WriteHeader(http.StatusInternalServerError)
 		return
 	}
-	if err := create(outPath, isoWorkDir, lca_api.BlockDeviceLabel); err != nil {
-		h.Log.WithError(err).Error("failed to create iso")
+	defer os.Remove(outPath)
```
carbonin, line 85: `tempFileName` doesn't actually write anything to `outPath` so while moving this here is fine, it doesn't actually change anything.

eranco74: Whenever `create` fails an empty tempFile leaks and stay in the `/data/iso-workdir/`

carbonin: Ah right, because create makes the file itself. Good catch

PR #170, opened 2024-12-10
MGMT-19487: Add documentation for cluster reinstall

carbonin:
Document how to reinstall a cluster using the resources from an existing install.

### DIFF
--- a/doc/reinstall.md
+++ b/doc/reinstall.md
@@ -29,7 +29,6 @@ Once the secrets are backed up some data needs to be removed to ensure they work
 
 Remove the following fields if they are present from all the previously created secret files:
 - metadata.creationTimestamp
-- metadata.labels (remove all except `image-based-installed.openshift.io/created`)
 - metadata.ownerReferences
 - metadata.resourceVersion
 - metadata.uid
@@ -86,13 +85,11 @@ type: Opaque
 
 #### Remove the existing cluster resources
 
-Delete the following resources associated with the original cluster installation:
+Ensure that the following resources associated with the original cluster installation have been deleted, or delete them manually:
 - `ClusterDeployment`
 - `ImageClusterInstall`
 - `BareMetalHost`
 - `DataImage`
-- `Secrets`
-- `ConfigMaps`
 
 #### Create the secrets from the original installation
 
--- a/doc/reinstall.md
+++ b/doc/reinstall.md
@@ -1,0 +1,118 @@
+# Cluster Reinstallation
+
+You can reinstall a cluster previously installed using the Image Based Install Operator on new hardware while preserving the cluster crypto, identity, and access.
+
+## Prerequisites
+
+- The cluster must have been originally installed using the Image Based Install Operator
+- The following secrets must still be present in the hub cluster in the same namespace as the cluster deployment
+    - `<clustername>-admin-kubeconfig`
+    - `<clustername>-admin-password`
+    - `<clustername>-seed-reconfiguration`
+- The new machine must have been installed with the desired seed image and ready for final reconfiguration
+
+## Reinstallation
+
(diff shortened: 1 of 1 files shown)


doc/reinstall.md
```diff
@@ -0,0 +1,120 @@
...
+
+You can trigger cluster installation by recreating the resources the same way they were created for the initial installation.
+
+Note:
+
+Some fields cannot be changed between install and reinstallation. These include:
+- `clusterDeployment.baseDomain`
+- `clusterDeployment.clusterName`
+- `imageClusterInstall.hostname`
+- `imageClusterInstall.nodeIP`
```
carbonin, line 115: Are these still true? Also maybe I should leave this note for when I implement the check for https://issues.redhat.com/browse/MGMT-19209?

carbonin: Looks like hostname might be relevant but node IP isn't actually used anywhere. I'll remove that one.

doc/reinstall.md
```diff
@@ -0,0 +1,119 @@
...
+#### Create the resources to install the cluster
+
+You can trigger cluster installation by recreating the resources the same way they were created for the initial installation.
+
+Note:
+
+Some fields cannot be changed between install and reinstallation. These include:
+- `clusterDeployment.baseDomain`
+- `clusterDeployment.clusterName`
+- `imageClusterInstall.hostname`
```
carbonin, line 114: I assume this is true, but can you confirm @eranco74 ?

doc/reinstall.md
```diff
@@ -0,0 +1,118 @@
...
+
+### Recreating the cluster
+
+#### Remove the existing cluster resources
+
+Delete the following resources associated with the original cluster installation:
+- `ClusterDeployment`
+- `ImageClusterInstall`
+- `BareMetalHost`
+- `DataImage`
```
sakhoury, line 93: Shouldn't the deletion of the `DataImage` object be handled by the BMO when the BMH object is deleted?

carbonin: I had some issues with this not being deleted so I added it here, but you're probably right in that _something_ is likely removing this during finalizer handling. Possibly even the ImageClusterInstall

doc/reinstall.md
```diff
@@ -0,0 +1,118 @@
...
+  name: ibi-test-seed-reconfiguration
+  namespace: ibi-test
+type: Opaque
+```
+
+### Recreating the cluster
+
+#### Remove the existing cluster resources
+
+Delete the following resources associated with the original cluster installation:
```
sakhoury, line 89:
```suggestion
Ensure that the following resources associated with the original cluster installation have been deleted, or delete them manually if they are still present:
```

doc/reinstall.md
sakhoury: Added a suggestion to reword the deletion.

tsorya: BMO will not delete it if you delete bmh, i think they have some bug with it, so probably better just to delete it

doc/reinstall.md
```diff
@@ -0,0 +1,118 @@
...
+oc get secret $CDNAME-seed-reconfiguration -oyaml > seed-reconfiguration.yaml
+```
+
+### Prepare the secrets for recreation
+
+Once the secrets are backed up some data needs to be removed to ensure they work correctly when recreated.
+
+Remove the following fields if they are present from all the previously created secret files:
+- metadata.creationTimestamp
+- metadata.labels (remove all except `image-based-installed.openshift.io/created`)
```
sakhoury, line 32: Should only the `image-based-installed.openshift.io/created` label be retained, or is it possible to keep the entire `metadata.labels` object?

sakhoury (approved):

PR #171, opened 2024-12-10
MGMT-19209: Validate important fields have not changed on reinstall

carbonin:
Allow most of the reconfig manifest to change based on the current data in the ClusterDeployment and ImageClusterInstall, but some fields must not change between install and reinstall.

This commit validates these fields match between the manifest generated based on the new config and the manifest in the secret from the original install.

Resolves https://issues.redhat.com/browse/MGMT-19209

### DIFF
--- a/internal/installer/reinstall.go
+++ b/internal/installer/reinstall.go
@@ -29,6 +29,10 @@ func (i *installer) WriteReinstallData(ctx context.Context, tmpWorkDir, isoWorkD
 		return fmt.Errorf("failed to decode local seed reconfiguration: %w", err)
 	}
 
+	if err := validateReinstallConfig(clusterConfig.Config, &secretSeedReconfig); err != nil {
+		return fmt.Errorf("cluster configuration is not valid for reinstall: %w", err)
+	}
+
 	clusterConfig.Config.KubeadminPasswordHash = secretSeedReconfig.KubeadminPasswordHash
 	clusterConfig.Config.KubeconfigCryptoRetention = secretSeedReconfig.KubeconfigCryptoRetention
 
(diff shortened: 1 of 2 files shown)


internal/installer/reinstall.go
```diff
@@ -49,6 +53,20 @@ func (i *installer) WriteReinstallData(ctx context.Context, tmpWorkDir, isoWorkD
...
 }
 
+func validateReinstall(newConfig, secretConfig *imagebased.SeedReconfiguration) error {
+	if newConfig.BaseDomain != secretConfig.BaseDomain {
+		return fmt.Errorf("provided base domain (%s) must match previous base domain (%s)", newConfig.BaseDomain, secretConfig.BaseDomain)
+	}
+	if newConfig.ClusterName != secretConfig.ClusterName {
+		return fmt.Errorf("provided cluster name (%s) must match previous cluster name (%s)", newConfig.ClusterName, secretConfig.ClusterName)
+	}
+	if newConfig.Hostname != secretConfig.Hostname {
```
tsorya, line 63: i wonder if hostname is important actually, it doesn't influence any key so why not to allow changing it?

carbonin: Wouldn't we need the same hostname for the existing kubeconfig to still work?

carbonin: Oh I guess that's a different dns entry. :shrug:

sakhoury (commented):
looks good!

internal/installer/reinstall_test.go
```diff
@@ -166,4 +166,26 @@ var _ = Describe("WriteReinstallData", func() {
...
+				Kubeconfig:        []byte("kubeconfig"),
+				KubeadminPassword: []byte("password"),
+				SeedReconfig:      []byte(data),
+			}
+
+			err = inst.WriteReinstallData(context.Background(), tmpWorkDir, isoWorkDir, idData)
+			Expect(err).To(HaveOccurred())
+			Expect(err.Error()).To(ContainSubstring(errSubstring))
+		},
+		Entry("check base domain has not changed", "base_domain", "example.org", "provided base domain (example.com) must match previous base domain (example.org)"),
```
eranco74, line 188: Add happy flow?

eranco74 (commented):
/lgtm

internal/installer/reinstall_test.go
carbonin: The tests above handle the happy flow. Everything is testing `WriteReinstallData`

sakhoury (approved):

PR #172, opened 2024-12-11
MGMT-19492: Add siteconfig secret label

carbonin:
This ensures that the siteconfig operator will backup and restore these secrets when it is orchestrating a reinstall for IBBF

Resolves https://issues.redhat.com/browse/MGMT-19492

Additionally this PR moves the labels and owner references to the mutate function in `createOrUpdateClusterCredentialSecret` so that we can ensure those are always set.


sakhoury (commented):
/lgtm

PR #195, opened 2025-01-17
MGMT-19745: Handle readOnlyRootFilesystem in pod deployments

carbonin:
MCE's installer is adding additional security features to our deployment when it is run.
We should also add those to our deployment here to catch any issues they might cause in presubmits rather than only hitting them when QE gets a build from MCE.

Specifically the addition of readOnlyRootFilesystem caused https://issues.redhat.com/browse/MGMT-19745. This would have been found in a presubmit if it was here before the relevant changes were made.

Also sets TMPDIR env var to /data. Some dependencies (openshift/installer) also use temp storage so it's better to set TMPDIR for the whole process rather than trying to enforce temp files be written to /data in code.

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -47,6 +47,8 @@ spec:
           value: "https"
         - name: MAX_CONCURRENT_RECONCILES
           value: "1"
+        - name: TMPDIR
+          value: /data
         securityContext:
           allowPrivilegeEscalation: false
           capabilities:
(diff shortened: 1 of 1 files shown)


carbonin:
Using https://github.com/stolostron/backplane-operator/blob/main/pkg/templates/charts/toggle/image-based-install-operator/templates/image-based-install-operator.yaml#L89 to make these changes.

PR #214, opened 2025-02-24
MGMT-20024: IBIO fails in case old dataImage exists and hangs during deletion.

eranco74:
The changes add a validation step in the ensureBMHDataImage method, the code now checks if a DataImage has a non-zero DeletionTimestamp and returns an explicit error if so.
Added a test case to verify that when a DataImage marked for deletion is processed, the controller sets the appropriate status condition and error message.

### DIFF
--- a/controllers/imageclusterinstall_controller_test.go
+++ b/controllers/imageclusterinstall_controller_test.go
@@ -922,6 +922,60 @@ var _ = Describe("Reconcile", func() {
 
 	})
 
+	It("Set status condition in case dataImage exists and being deleted", func() {
+		bmh := bmhInState(bmh_v1alpha1.StateAvailable)
+		bmh.Spec.ExternallyProvisioned = true
+		Expect(c.Create(ctx, bmh)).To(Succeed())
+
+		clusterInstall.Spec.BareMetalHostRef = &v1alpha1.BareMetalHostReference{
+			Name:      bmh.Name,
+			Namespace: bmh.Namespace,
+		}
+		Expect(c.Create(ctx, clusterInstall)).To(Succeed())
+		Expect(c.Create(ctx, clusterDeployment)).To(Succeed())
+
+		dataImage := bmh_v1alpha1.DataImage{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:       bmh.Name,
+				Namespace:  bmh.Namespace,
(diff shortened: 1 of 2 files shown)


controllers/imageclusterinstall_controller_test.go
```diff
@@ -922,6 +922,65 @@ var _ = Describe("Reconcile", func() {
...
+				Namespace:  bmh.Namespace,
+				Finalizers: []string{"test.finalizer"}, // Prevent immediate deletion
+			},
+		}
+		Expect(c.Create(ctx, &dataImage)).To(Succeed())
+
+		defer func() {
+			dataImage.SetFinalizers([]string{}) // Remove finalizer
+			Expect(c.Update(ctx, &dataImage)).To(Succeed())
+		}()
```
carbonin, line 949: Should we do this in an `After` block of some sort?
Separately is this really part of the test or some kind of cleanup?
If it's cleanup I feel like it's just not necessary since we're re-initializing the fake client for each test.

controllers/imageclusterinstall_controller_test.go
```diff
@@ -922,6 +922,65 @@ var _ = Describe("Reconcile", func() {
...
+
+		Expect(c.Delete(ctx, &dataImage)).To(Succeed())
+		Eventually(func() *metav1.Time {
+			err := c.Get(ctx, client.ObjectKey{Name: dataImage.Name, Namespace: dataImage.Namespace}, &dataImage)
+			if err != nil {
+				return nil
+			}
+			return dataImage.DeletionTimestamp
+		}).ShouldNot(BeNil()) // Ensure deletion timestamp is set
```
carbonin, line 958: I wasn't aware that this is how this worked.

I mean I guess it does, but I thought the fake client delete just removed the object rather than being aware of finalizers and such.

carbonin (commented):
Looks fine, just some questions about the test.

controllers/imageclusterinstall_controller_test.go
eranco74: I know, that's why I added this check, to make sure the fake client is aware of the finalizer and the object isn't deleted.
At first I tried to set the DeletionTimestamp but I got:
`Unable to edit test-bmh: metadata.deletionTimestamp field is immutable`
So had to come up with this...

PR #217, opened 2025-02-27
[backplane-2.8] MGMT-20069: IBIO fails in case old dataImage exists and hangs during deletion

eranco74:
In case dataImage exists make sure it's not marked for deletion, in case DeletionTimestamp is set update the ClusterInstallRequirementsMet condition with an informative message
Signed-off-by: Eran Cohen <eranco@redhat.com>

### DIFF
--- a/controllers/imageclusterinstall_controller.go
+++ b/controllers/imageclusterinstall_controller.go
@@ -19,7 +19,6 @@ package controllers
 import (
 	"bytes"
 	"context"
-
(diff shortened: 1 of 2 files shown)


eranco74:
```
{"component":"clonerefs","error":"1 clone records failed","file":"sigs.k8s.io/prow/cmd/clonerefs/main.go:40","func":"main.main","level":"fatal","msg":"Failed to clone refs","severity":"fatal","time":"2025-02-27T19:05:36Z"}
```
/retest

PR #233, opened 2025-03-16
MGMT-20128: Image based install operator not reconciling after pod is  restarted

eranco74:
The monitor will list existing ImageClusterInstall CRs and call client.Status.Update for all ICI that didn't complete the installation.
This should enqueue the ICI CRs for reconciliation when the controller starts

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -17,6 +17,7 @@ limitations under the License.
 package main
 
 import (
+	"context"
 	"flag"
 	"fmt"
 	"net/http"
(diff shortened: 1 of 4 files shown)


carbonin:
@trewest How urgent is this fix?
We've been told that something like this should be happening automatically just by using the operator framework and I'd like to more deeply investigate why that doesn't seem to be the case in our controller, but I don't have the time to look into it now. Do you feel like we should merge this in the mean-time?

carbonin (approved):
I haven't had a chance to investigate this any further so I'm feeling like I won't get that chance for a while. Might as well merge this to address the issue until I do get that chance.

PR #234, opened 2025-03-18
MGMT-20024: fix the hot-loop in case dataImage hang while deleting

eranco74:
Updates error handling to manage DataImage in deletion process better.
When the deletion timestamp is non-zero, a warning is logged, the status is updated, and a requeueAfter result with a one-minute delay is returned. 
This check was moved out of `ensureBMHDataImage` method, to allow early and specific handling for such case since it may conflict with the  requirementsMet status condition set by validateBMH method.
Additionally, the return type of the validateBMH method is simplified to streamline error handling.

### DIFF
--- a/controllers/imageclusterinstall_controller.go
+++ b/controllers/imageclusterinstall_controller.go
@@ -221,17 +221,17 @@ func (r *ImageClusterInstallReconciler) Reconcile(ctx context.Context, req ctrl.
 		}
 	}
 
-	res, err := r.validateBMH(ctx, log, ici, bmh)
-	if err != nil || !res.IsZero() {
-		return res, err
+	if err = r.validateBMH(ctx, log, ici, bmh); err != nil {
+		// in case of failure to validate the BMH just requeue
+		return ctrl.Result{RequeueAfter: 30 * time.Second}, nil
 	}
 
-	if err := r.setClusterInstallMetadata(ctx, log, ici, clusterDeployment); err != nil {
+	if err = r.setClusterInstallMetadata(ctx, log, ici, clusterDeployment); err != nil {
 		log.WithError(err).Error("failed to set ImageClusterInstall data")
 		return ctrl.Result{}, err
 	}
 
-	res, err = r.writeInputData(ctx, log, ici, clusterDeployment, bmh)
+	res, err := r.writeInputData(ctx, log, ici, clusterDeployment, bmh)
 	if !res.IsZero() || err != nil {
 		if err != nil {
 			if updateErr := r.setImageReadyCondition(ctx, ici, err); updateErr != nil {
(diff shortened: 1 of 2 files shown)


controllers/imageclusterinstall_controller.go
```diff
@@ -259,8 +259,16 @@ func (r *ImageClusterInstallReconciler) Reconcile(ctx context.Context, req ctrl.
 		if updateErr := r.setHostConfiguredCondition(ctx, ici, err); updateErr != nil {
 			log.WithError(updateErr).Error("failed to create DataImage")
 		}
-		return ctrl.Result{}, err
```
carbonin, line 262: Why remove this? Do we not want to fail here?

eranco74: We do, thanks

carbonin (approved):

controllers/imageclusterinstall_controller.go
```diff
@@ -220,18 +220,28 @@ func (r *ImageClusterInstallReconciler) Reconcile(ctx context.Context, req ctrl.
 			return ctrl.Result{}, err
 		}
 	}
+	// This code was moved here to avoid updating the RequirementsMet condition back and forth
```
carbonin, line 223: If you want to keep this comment I'd avoid talking about how it has been "moved". This feels like it belongs more in the commit message

Maybe something like
```suggestion
	// validate the data image isn't being deleted before setting `RequirementsMet` to success in `validateBMH`
```

controllers/imageclusterinstall_controller.go
```diff
@@ -220,18 +220,28 @@ func (r *ImageClusterInstallReconciler) Reconcile(ctx context.Context, req ctrl.
...
+	if err = r.setClusterInstallMetadata(ctx, log, ici, clusterDeployment); err != nil {
 		log.WithError(err).Error("failed to set ImageClusterInstall data")
 		return ctrl.Result{}, err
 	}
 
-	res, err = r.writeInputData(ctx, log, ici, clusterDeployment, bmh)
+	res, err := r.writeInputData(ctx, log, ici, clusterDeployment, bmh)
 	if !res.IsZero() || err != nil {
 		if err != nil {
 			if updateErr := r.setImageReadyCondition(ctx, ici, err); updateErr != nil {
```
carbonin, line 247: If this is setting `RequirementsMet` don't we have the same issue here (and anywhere else that sets that condition after the call to `validateBMH`, like `setHostConfiguredCondition`)?

carbonin (commented):
Looking at how many functions could set `RequirementsMet` back to false after `validateBMH` sets it to true I'm wondering if the better solution wouldn't be to move the success update out of `validateBMH` to after all the other potential error cases?

controllers/imageclusterinstall_controller.go
```diff
@@ -261,6 +261,10 @@ func (r *ImageClusterInstallReconciler) Reconcile(ctx context.Context, req ctrl.
 		}
 		return ctrl.Result{}, err
 	}
+	if updateErr := r.setRequirementsMetCondition(ctx, ici, corev1.ConditionTrue, v1alpha1.HostValidationSucceeded, v1alpha1.HostValidationsOKMsg); updateErr != nil {
```
carbonin, line 264: I think this should still be after the call to `setHostConfiguredCondition` in the error case for `updateBMHProvisioningState`, right?

eranco74: Hmm, I think I rather fix `updateBMHProvisioningState` the only error case there is failure to path the BMH with an annotation, that isn't related to the requirementsMet condition.

eranco74:
devscripts failed
/retest

PR #235, opened 2025-03-18
MGMT-20069: fix the hot-loop in case dataImage hang while deleting

eranco74:


carbonin:
This is made against the wrong branch so closing in favor of https://github.com/openshift/image-based-install-operator/pull/247

PR #267, opened 2025-05-01
Add required parameters for konflux enterprise contract

carbonin:


Kyl-Bempah:
There shouldn't be a need to add the these values you can just edit the existing values here:
- https://github.com/openshift/image-based-install-operator/pull/267/files#diff-a547f024da351d3f7e788fb89dba094b6862c6dde3a124e0f244e80531ed2113L98
- https://github.com/openshift/image-based-install-operator/pull/267/files#diff-a53177cb9866d779334c3b9b64b1d8993f304177b38d9e35eebc063b34c14f0eL95

PR #270, opened 2025-05-05
MGMT-20179: Fix IBIO status conditions

zszabo-rh:
Refactoring Reconcile() in relation with MGMT-20179

Reconcile sets the following reasons when requirements not met due to some failure:
- ConfigurationNotReadyReason: BMH or CD reference missing or the resource is not available
- HostValidationPending: BMH provisioning is not ready yet
- HostValidationFailedReason: 
- ImageNotReadyReason: 
- HostConfigurationPendingReason
- HostConfigurationFailedReason

And finally when everything goes fine during reconcile and RequirementsMet condition is set:
- HostConfigurationSucceededReason


controllers/imageclusterinstall_controller.go
```diff
@@ -246,42 +244,44 @@ func (r *ImageClusterInstallReconciler) Reconcile(ctx context.Context, req ctrl.
...
+	cond.Reason = v1alpha1.HostConfigurationFailedReason
+
+	dataImage, res, err := r.ensureBMHDataImage(ctx, log, bmh, imageUrl)
 	if err != nil {
-		log.WithError(err).Error("failed to create BareMetalHost DataImage")
-		if updateErr := r.setHostConfiguredCondition(ctx, ici, err); updateErr != nil {
-			log.WithError(updateErr).Error("failed to create DataImage")
+		cond.Message = "failed to create BareMetalHost DataImage"
+		if !res.IsZero() {
+			cond.Reason = v1alpha1.HostConfigurationPendingReason
```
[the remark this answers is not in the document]
zszabo-rh: since it's the infamous _previous dataImage is being deleted_ scenario..
in this case we gonna requeu with 30s timer, so I assumed "pending" fits better

controllers/imageclusterinstall_controller.go
```diff
@@ -246,42 +244,44 @@ func (r *ImageClusterInstallReconciler) Reconcile(ctx context.Context, req ctrl.
...
-		}
+		cond.Message = "failed to update BareMetalHost provisioning state"
+		log.WithError(err).Error(cond.Message)
 		return ctrl.Result{}, err
 	}
 	if !annotationExists(&bmh.ObjectMeta, ibioManagedBMH) {
 		// TODO: consider replacing this condition with `dataDisk.Status.AttachedImage`
-		log.Infof("Nothing to do, waiting for BMH to get %s annotation", ibioManagedBMH)
+		cond.Reason = v1alpha1.HostConfigurationPendingReason
+		cond.Message = fmt.Sprintf("Waiting for BMH to get %s annotation", ibioManagedBMH)
```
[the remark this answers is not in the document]
zszabo-rh: Changed to:
`cond.Message = fmt.Sprintf("Waiting for BMH provisioning state to be StateAvailable or StateExternallyProvisioned, current state is: %s", bmh.Status.Provisioning.State)
`

controllers/imageclusterinstall_controller.go
```diff
@@ -152,136 +154,230 @@ func (r *ImageClusterInstallReconciler) Reconcile(ctx context.Context, req ctrl.
...
+		log.Infof("Setting BareMetalHost (%s/%s) ExternallyProvisioned spec", bmh.Namespace, bmh.Name)
+		patch := client.MergeFrom(bmh.DeepCopy())
+		bmh.Spec.ExternallyProvisioned = true
+		if err := r.Patch(ctx, bmh, patch); err != nil {
+			return ctrl.Result{}, err
+		}
+	
+	}
+
+	if err := r.setClusterInstallMetadata(ctx, log, ici, cd); err != nil {
```
carbonin, line 298: This is a bit odd to have in a function called "validateHost". I wouldn't expect these kinds of side effects from a validation. Maybe it doesn't belong here or maybe we can come up with a better name for the function.

controllers/imageclusterinstall_controller.go
```diff
@@ -152,136 +154,230 @@ func (r *ImageClusterInstallReconciler) Reconcile(ctx context.Context, req ctrl.
...
+			cond.Reason = v1alpha1.ImageCreationFailedReason
+			cond.Message = "failed to create image"
+			log.WithError(err).Error(cond.Message)
 		}
-		return res, err
+		return "", res, err
 	}
 
-	r.labelReferencedObjectsForBackup(ctx, log, ici, clusterDeployment)
+	r.labelReferencedObjectsForBackup(ctx, log, ici, cd)
```
carbonin, line 330: This isn't really part of the image creation, can this maybe just live in the caller since it's never going to cause a condition failure?

carbonin (commented):
Let's keep messages formatted consistently. Things like either capitalizing the first word or not I think should always be the same.

controllers/imageclusterinstall_controller.go
zszabo-rh: yes it's just a mid-phase step that doesn't really belong to host validation nor image creation
I've moved it back to caller, but do you think it's still ok to use the `HostValidationFailedReason` for this? I can even introduce a dedicated phase and condition reason just for this single step (maybe also bundle together with `labelReferencedObjectsForBackup`), not sure which is better.

carbonin: I think it's fine either way

controllers/imageclusterinstall_controller.go
```diff
@@ -152,136 +154,230 @@ func (r *ImageClusterInstallReconciler) Reconcile(ctx context.Context, req ctrl.
...
+	return filepath.Join(namespacesDir, namespace, uid, FilesDir, ClusterConfigDir)
+}
+
+func (r *ImageClusterInstallReconciler) validateConfiguration(
+	ctx context.Context,
+	ici *v1alpha1.ImageClusterInstall,
+	cond *hivev1.ClusterInstallCondition,
+	log logrus.FieldLogger,
+) (*hivev1.ClusterDeployment, *bmh_v1alpha1.BareMetalHost, error) {
+	cond.Reason = v1alpha1.ConfigurationPendingReason
```
carbonin, line 240: Given that this is acting as a kind of state transition, and that the caller is responsible for actually saving the condition can we move these to the caller as well? I think it would make it a bit more clear what's going on.

Actually if all of the condition changes could be done in the caller I think that would be better, but I understand if you don't want to create a ton of return values given that (at least this function) changes both the message and reason in some error cases.

zszabo-rh: ok I moved the state defaults to the caller, but yes, it would require a lot of extra work to identify each and every error case and do the corresponding cond change also there, not sure if worth the effort.

zszabo-rh:
/hold
analyzing e2e logs

controllers/imageclusterinstall_controller_test.go
```diff
@@ -1959,7 +1960,8 @@ var _ = Describe("Reconcile with DataImageCoolDownPeriod set to 1 second", func(
 		}
 		installerSuccess()
 		res, err := r.Reconcile(ctx, req)
-		Expect(err).NotTo(HaveOccurred())
+		Expect(err).To(HaveOccurred())
```
carbonin, line 1963: Should this really be an error?

controllers/imageclusterinstall_controller_test.go
```diff
@@ -1518,7 +1519,7 @@ var _ = Describe("Reconcile", func() {
 		Expect(infoOut.MachineNetwork[0].CIDR.String()).To(Equal(clusterInstall.Spec.MachineNetwork))
 	})
 
-	It("in case there is no actual bmh under the reference we should not return error", func() {
+	It("in case there is no actual bmh under the reference we should return error", func() {
```
carbonin, line 1522: Do we really need both this test and `Set ClusterInstallRequirementsMet to false in case there is not actual bmh under the reference`?

controllers/imageclusterinstall_controller_test.go
```diff
@@ -1295,21 +1296,21 @@ var _ = Describe("Reconcile", func() {
...
+	It("sets the ClusterInstallRequirementsMet condition to false when ClusterDeploymentRef is unset", func() {
 		clusterInstall.Spec.ClusterDeploymentRef = nil
 		Expect(c.Create(ctx, clusterInstall)).To(Succeed())
 		key := types.NamespacedName{
 			Namespace: clusterInstallNamespace,
 			Name:      clusterInstallName,
 		}
 		res, err := r.Reconcile(ctx, ctrl.Request{NamespacedName: key})
-		Expect(err).NotTo(HaveOccurred())
+		Expect(err).To(HaveOccurred())
```
carbonin, line 1307: This is a strange behavior change, why would we want an error in this case? An error will cause an additional reconcile where we would expect to only reconcile once the resource was changed.

controllers/imageclusterinstall_controller_test.go
```diff
@@ -997,7 +997,8 @@ var _ = Describe("Reconcile", func() {
 		}
 		installerSuccess()
 		res, err := r.Reconcile(ctx, req)
-		Expect(err).NotTo(HaveOccurred())
+		Expect(err).To(HaveOccurred())
+		Expect(err.Error()).To(ContainSubstring("current state is: registering"))
```
carbonin, line 1001: Why error here?

carbonin (commented):
Looking at the tests I'm a bit worried about the amount of new error cases coming from `Reconcile`. Returning an error incurs more reconcile calls and I'm not sure that's what we want in all these cases.

controllers/imageclusterinstall_controller_test.go
zszabo-rh: tbh this rather crucial information was new for me, somehow I assumed setting an error is more like an easily accessible debug info for the user, but otherwise does not affect the reconcile loop.

Thanks for clarifying this, I refactored the errors to limit them only for real issues.

What I don't like now is that basically each "main phase method" called by Reconcile() returns values by unique logic. Do you think I should make them uniform, like have them all return a clear statement if reconcile should continue or not (even if this can be easily figured out by checking the rest of the values)? I feel that approach a bit forced without having much benefits, but I'll consider again if you think otherwise.

carbonin (commented):
I see what you mean about the error handling. If you want to try for a generic approach we did something like this in the [bmh agent controller in assisted-service](https://github.com/openshift/assisted-service/blob/6ef17dd3ff8304682d3b122ad769bd69c5182ccf/internal/controller/controllers/bmh_agent_controller.go#L136-L149), but I'm still not sure I like it.

Maybe it would be useful here though with some tweaks (like not using a type called `reconcileComplete` where you're not stopping reconcile :tired_face: )

carbonin:
Is there a reason this is still on hold?

Feature freeze is today for 2.9 if we want to get this in ...

PR #335, opened 2025-09-15
MGMT-21758: Update konflux dockerfile

carbonin:
Changes from https://github.com/openshift/image-based-install-operator/commit/5ee89af3803f276669a99e7c4af033d1f9e5baad need to be pulled into the konflux dockerfile.

cc @danmanor


carbonin:
Starting with just the go version to see if this passes. That will unblock builds, then I'll figure out how to update the rpm lock file.

danmanor (approved):

PR #798, opened 2026-03-31
[backplane-2.17] ACM-31897: Use the common konflux pipeline

giladravid16:


gparvin:
/lgtm

We are looking for this fix to unblock the MCE 2.17 trusted pipeline.  Thanks

PR #801, opened 2026-04-19
ACM-30179: Use TLS configuration from the APIServer when available

giladravid16:
OCP 4.22 indroduces a new feature in the APIServer for configuring TLS configuration (min TLS version and ciphers), and we want our components to also use the same TLS configuration that is defined in the APIServer.

When starting a components we will attempt to get the APIServer and check its TLSAdherencePolicy (decides whether we should apply the TLS configuration) and TLSProfileSpec (the actual TLS configuration).
For our operator we also add a controller called SecurityProfileWatcher (provided by openshift/controller-runtime-common) in order to force a restart when these values change.
For our server we implement a similar behavior ourselves.

This also required updating a lot of packages including the client-go which further required setting the environment variable KUBE_FEATURE_WatchListClient to false on the operator in order to fix a known bug where controllers don't start due to a cache sync issue.


cmd/server/main.go
```diff
@@ -75,3 +112,43 @@ func main() {
...
+		}
+		updated, ok := event.Object.(*configv1.APIServer)
+		if !ok {
+			continue
+		}
+
+		if current.TLSAdherencePolicy != updated.Spec.TLSAdherence {
+			log.Infof("TLS adherence policy has changed, shutting down to reload, oldPolicy: %v, newPolicy: %v",
+				current.TLSAdherencePolicy, updated.Spec.TLSAdherence)
+			os.Exit(0)
```
carbonin: Does this trigger the graceful server shutdown?

config/manager/manager.yaml
```diff
@@ -141,3 +148,18 @@ spec:
     name: config-server
   selector:
     app: image-based-install-operator
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: image-based-install-metrics
```
carbonin: If we didn't even have a metrics service before can we just stop serving metrics rather than working to expose a TLS endpoint that we don't care about?

carbonin (commented):
For my own curiosity, could you include something (even just a reference) about exactly what this changes? What do `TLSAdherencePolicy` and `TLSProfileSpec` actually indicate? Why would someone want to use them?

Also given that we're using service serving certificates to actually generate the certs, wouldn't whatever cluster operator (network operator?) is generating those already create certs that comply with whatever policy is required?

rccrdpccl:
@giladravid16 in general it would be great if huge PRs as this one would split the vendor change into a separate commit, so that we  can be easily review the actual change

internal/tlsconfig/tlsconfig.go
```diff
@@ -0,0 +1,81 @@
...
+
+	var tlsConfig func(*tls.Config)
+
+	if libgocrypto.ShouldHonorClusterTLSProfile(tlsAdherencePolicy) {
+		profileTLSConfig, unsupportedCiphers := crtls.NewTLSConfigFromProfile(tlsProfileSpec)
+		if len(unsupportedCiphers) > 0 {
+			log.Info("TLS configuration contains unsupported ciphers that will be ignored", "unsupportedCiphers", unsupportedCiphers)
+		}
+		tlsConfig = profileTLSConfig
+	} else {
```
rccrdpccl: nit: else condition is never necessary

internal/tlsconfig/tlsconfig.go
```diff
@@ -0,0 +1,81 @@
...
+		defaultTLSConfig, unsupportedCiphers := crtls.NewTLSConfigFromProfile(*configv1.TLSProfiles[libgocrypto.DefaultTLSProfileType])
+		if len(unsupportedCiphers) > 0 {
+			log.Info("TLS configuration contains unsupported ciphers that will be ignored", "unsupportedCiphers", unsupportedCiphers)
+		}
+		tlsConfig = defaultTLSConfig
+	}
```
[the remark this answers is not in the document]
rccrdpccl: I think the rabbit has a point

cmd/server/main.go
giladravid16: I updated it so it does now

config/manager/manager.yaml
giladravid16: Right now we don't create a service for metrics, but it's exposted via http. Users can create a service for it right now if they wanted to. Do you prefer disabling it? Or leaving it as http?

internal/tlsconfig/tlsconfig.go
giladravid16: How come? When tlsAdherence is not StrictAllComponents then we ignore what's in the tlsProfileSpec and use the default instead

rccrdpccl: It's not about the logic but the style. Else condition makes the code more difficult to read.
To remove it in this case you could for instance extract this logic to a function and return early in one of the two cases

in any case it's just nitpicking

giladravid16: oh ok, yeah you're right

config/manager/manager.yaml
carbonin: Do we have a real reason to keep it? Are we exposing any meaningful metrics? If not then I'd say just disable it.

giladravid16: I disabled it for now.
We don't add any metrics of our own so we only export the default metrics. The list is pretty long, but it's mostly:
1. resource usage (like cpu/memory/network)
2. go metrics (like number of goroutines)
3. controller/webhook metrics (like number of reconciles and admission requests)

rccrdpccl:
LGTM

giladravid16:
@carbonin I added in the description and the commit that the TLS configuration includes the min TLS version and the ciphers that can be used. The idea is that eventually all components in the cluster will use the same TLS configuration that users will be able to define in the APIServer. The actual TLS configuration is defined in the `TLSProfileSpec` and users can decide which components need to adheare to the configuration with the `TLSAdherencePolicy` field.

carbonin (approved):

PR #812, opened 2026-04-23
ACM-33186: Revert Renovate configuration for Hive updates

shay23bra:
Revert the Renovate/MintMaker configuration that enables hive/apis digest tracking.
MintMaker generates bare commit hashes instead of proper Go pseudo-versions,
which breaks go.mod.

Reverts changes from [ACM-29247](https://redhat.atlassian.net/browse/ACM-29247) / [ACM-29261](https://redhat.atlassian.net/browse/ACM-29261) / ACM-29244.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -9,29 +9,11 @@
     "prConcurrentLimit": 0,
     "enabledManagers": [
         "custom.regex",
-        "tekton",
-        "gomod"
+        "tekton"
     ],
     "tekton": {
         "managerFilePatterns": [
             "/^.tekton/*/"
-        ]
-    },
-    "gomod": {
-        "postUpdateOptions": [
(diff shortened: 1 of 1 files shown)


carbonin:
I don't really understand. What about this is broken?

PRs like https://github.com/openshift/image-based-install-operator/pull/813 wouldn't merge if the dependencies were broken. Would the image fail to build?

shay23bra:
> I don't really understand. What about this is broken?
> 
> PRs like #813 wouldn't merge if the dependencies were broken. Would the image fail to build?

@carbonin There's an issue with Renovate, it's inconsistent and opening too many PRs with issues accross all our repos, so it's just until the fix is released. You are right, if there's an issue it will not be merged but there are too many of them.

gamli75 (approved):

